PR #7, opened 2015-05-02
pos_debt_notebook: Port to new API

alexis-via:
Add debt field in account.journal form view
Add debt field in res.partner kanban view
Add POT files and FR translation
Code style + PEP8

### DIFF
--- a/pos_debt_notebook/models.py
+++ b/pos_debt_notebook/models.py
@@ -1,52 +1,52 @@
- # -*- coding: utf-8 -*-
-from openerp.osv import osv,fields
-from openerp import SUPERUSER_ID
+# -*- coding: utf-8 -*-
+from openerp import models, fields, api
+import openerp.addons.decimal_precision as dp
 
-class res_partner(osv.Model):
+
+class ResPartner(models.Model):
     _inherit = 'res.partner'
 
-    def _get_debt(self, cr, uid, ids, field_name, arg, context=None):
-        obj_data = self.pool.get('ir.model.data')
-        debt_account_id = obj_data.get_object_reference(cr, uid, 'pos_debt_notebook', 'debt_account')[1]
-        debt_journal_id = obj_data.get_object_reference(cr, uid, 'pos_debt_notebook', 'debt_journal')[1]
+    @api.multi
+    def _get_debt(self):
+        debt_account = self.env.ref('pos_debt_notebook.debt_account')
+        debt_journal = self.env.ref('pos_debt_notebook.debt_journal')
 
-        cr.execute("""SELECT l.partner_id, SUM(l.debit-l.credit)
-                      FROM account_move_line l
-                      WHERE l.account_id = %s
-                      AND l.partner_id IN %s
-                      GROUP BY l.partner_id
-                      """,
-                   (debt_account_id, tuple(ids),))
+        self._cr.execute(
+            """SELECT l.partner_id, SUM(l.debit - l.credit)
+            FROM account_move_line l
+            WHERE l.account_id = %s AND l.partner_id IN %s
+            GROUP BY l.partner_id
+            """,
+            (debt_account.id, tuple(self.ids)))
 
         res = {}
-        for id in ids:
-            res[id] = 0
-        for id,val in cr.fetchall():
-            res[id] = val or 0
+        for partner in self:
+            res[partner.id] = 0
+        for partner_id, val in self._cr.fetchall():
+            res[partner_id] += val
 
-        statement_obj = self.pool.get('account.bank.statement')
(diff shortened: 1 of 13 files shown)


yelizariev:
@alexis-via thank you for contribution. I'll review update soon. 
Could you please create changlog file according to instruction below and update versions to 1.0.1:

Updates: Users who have installed your modules in Odoo 8.0 will see update notifications in their Settings/Updates menu if you increment the version number of any module. If they choose to update, the new module version will be installed and their server will be restarted, but the module data will not be re-synchronized automatically. You should avoid non-backwards compatible changes such as DB schema or API changes in stable series! If you have a doc/changelog.rst file inside your module, its contents will be displayed along with the notification updates.

tg_pos_debt_notebook/static/src/js/tg_pos.js
```diff
@@ -16,7 +16,7 @@ function tg_pos_debt_notebook(instance, module){ //module is instance.point_of_s
                     self.partners = partners;
                     self.db.add_partners(partners);
```
yelizariev: By the way, we can update it to make it real inheritance rather than overwritting.
You can check as I done it in tg_pos_packs/static/src/js/pos.js
So, it should be something like this:

```
             loaded.then(function(){
                 return self.fetch('res.partner', ['debt']);
             }).then(function(partners){
            $.each(partners, function(){
                $.extend(self.db.get_partner_by_id(this.id) || {}, this)
            });
            return $.when();
```

yelizariev:
The line at tg_pos_debt_notebook/README.rst:

```
Tested on Odoo 8.0 d023c079ed86468436f25da613bf486a4a17d625
```

Should be updated or removed.
I thinks it's better to have such note. So, the same line should be added to pos_debt_notebook/README.rst

pos_debt_notebook/models.py
```diff
@@ -1,52 +1,52 @@
...
     _inherit = 'res.partner'
 
-    def _get_debt(self, cr, uid, ids, field_name, arg, context=None):
-        obj_data = self.pool.get('ir.model.data')
-        debt_account_id = obj_data.get_object_reference(cr, uid, 'pos_debt_notebook', 'debt_account')[1]
-        debt_journal_id = obj_data.get_object_reference(cr, uid, 'pos_debt_notebook', 'debt_journal')[1]
+    @api.multi
+    def _get_debt(self):
+        debt_account = self.env.ref('pos_debt_notebook.debt_account')
+        debt_journal = self.env.ref('pos_debt_notebook.debt_journal')
```
bealdav, line 12: Could be

``` python
self.env.ref('debt_account')    
# instead of
self.env.ref('pos_debt_notebook.debt_account')
```

module name is implicit

alexis-via, line 12: @bealdav I tried your suggestion, but it crashes :

```
 File "/home/alexis/new_boite/dev/pos-yelizariev/pos_debt_notebook/models.py", line 11, in _get_debt
    debt_account = self.env.ref('debt_account')
  File "/home/alexis/new_boite/dev/odoo/openerp/api.py", line 737, in ref
    return self['ir.model.data'].xmlid_to_object(xml_id, raise_if_not_found=raise_if_not_found)
  File "/home/alexis/new_boite/dev/odoo/openerp/api.py", line 239, in wrapper
    return new_api(self, *args, **kwargs)
  File "/home/alexis/new_boite/dev/odoo/openerp/api.py", line 462, in new_api
    result = method(self._model, cr, uid, *args, **kwargs)
  File "/home/alexis/new_boite/dev/odoo/openerp/addons/base/ir/ir_model.py", line 950, in xmlid_to_object
    t = self.xmlid_to_res_model_res_id(cr, uid, xmlid, raise_if_not_found)
  File "/home/alexis/new_boite/dev/odoo/openerp/api.py", line 241, in wrapper
    return old_api(self, *args, **kwargs)
  File "/home/alexis/new_boite/dev/odoo/openerp/addons/base/ir/ir_model.py", line 936, in xmlid_to_res_model_res_id
    return self.xmlid_lookup(cr, uid, xmlid)[1:3]
  File "/home/alexis/new_boite/dev/odoo/openerp/api.py", line 241, in wrapper
    return old_api(self, *args, **kwargs)
  File "<string>", line 2, in xmlid_lookup
  File "/home/alexis/new_boite/dev/odoo/openerp/tools/cache.py", line 74, in lookup
    value = d[key] = self.method(*args, **kwargs)
  File "/home/alexis/new_boite/dev/odoo/openerp/addons/base/ir/ir_model.py", line 923, in xmlid_lookup
    module, name = xmlid.split('.', 1)
ValueError: need more than 1 value to unpack
```

So let's keep the code as it is.

PR #11, opened 2015-06-12
8.0 tg_pos_debt_notebook: Add button "pay full debt" 

alexis-via:
Add button "pay full debt" to make it easier for the user to manage the scenario where a customer comes to pay his debt

Enhance/cleanup JS code
Declare CSS file in XML, as done in other modules
Update POT file and FR translation

### DIFF
--- a/pos_debt_notebook/data.xml
+++ b/pos_debt_notebook/data.xml
@@ -33,5 +33,20 @@
             <field name="journal_ids" eval="[(4, ref('pos_debt_notebook.debt_journal'))]" />
         </record>
 
+        <record id="product_pay_debt" model="product.product">
+            <field name="name">Pay Debt</field>
+            <field name="default_code">POS-DEBT</field>
+            <field name="sale_ok" eval="True"/>
+            <field name="available_in_pos" eval="True"/>
+            <field name="list_price">0</field>
+            <field name="type">service</field>
+            <field name="uom_id" ref="product.product_uom_unit"/>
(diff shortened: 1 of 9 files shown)


yelizariev:
There is small issue.
Debt is not updated after clicking "Full debt Button"

selection_001
It is updated only after adding new line
selection_002

yelizariev:
I suggest to show current Debt inside Button:

```
Pay Full Debt
(4813.00)
```

Or maybe change title to:

```
Current debt: 4813.00
```

yelizariev:
I am sorry, I didn't notice Debt value at Tab. So, we don't need debt value at button

alexis-via:
@yelizariev Thanks for your review. I also noticied that the debt amount is only updated when a new payment line is added... I'll try to find the technique to do that. If someone has an idea on the proper technique to do that, it would be very much appreciated.

yelizariev:
Alexis,
thank you for your work.

Could you fix css issue
odoo - google chrome_001

yelizariev:
I think button "Select Customer and pay full debt" can be located near Debt field.

Also, I suggest to hide Pay debt buttons when customer don't have debts

yelizariev:
Alexis,
thank you for new commits.

Could you please update version and add changlelog to modules.

yelizariev:
@alexis-via I'd like to merge your update. Do you want to add something else and need some time? If not, could you please set version to '1.0.2' in __openerp__.py file and add short notes in doc/changelog.rst
Also, if you like, feel free to add your name at author list or at module description.

PR #12, opened 2015-06-30
FIX compiled python files problem

draugiskisprendimai:
Add ignore compiled python files


yelizariev:
Hello
thank you for you PR. Unfortunately, I'd like to close it, because I believe, that such kind of rules for gitignore should be done in your local git config

PR #21, opened 2015-10-01
Fixed pos_product_available product counter when create invoice

almas:
Hi bug fixed. Thanks for greate module.

### DIFF
--- a/pos_product_available/static/src/js/pos.js
+++ b/pos_product_available/static/src/js/pos.js
@@ -46,21 +46,19 @@ function pos_product_available(instance, module){
             var invoiced = PosModelSuper.prototype.push_and_invoice_order.call(this, order);
 
             if (order && order.get_client()){
-                console.log(order);
-                try {
-                    order.orderlines.each(function(line){
-                        var product = line.get_product();
-                        product.qty_available -= line.get_quantity();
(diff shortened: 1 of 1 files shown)


pos_product_available/static/src/js/pos.js
```diff
@@ -40,7 +40,32 @@ function pos_product_available(instance, module){
...
+                }
+                catch(err) {
+                    order.get("orderLines").each(function(line){
+                        var product = line.get_product();
+                        product.qty_available -= line.get_quantity();
+                        self.refresh_qty_available(product);
+                    })
+                }
+                
+            }
```
yelizariev, line 63: Some notes:
- Could you explain why do we need to check  order.get_client()?
- Please remove console.log
- I suggest to manually check version rather than use try-catch block
  
  if (order.orderlines){
      ...
  } else if (order.get("orderLines")){
      ...
  }

yelizariev:
You are right,
I didn't test  it with "invoice" button.

Could you check my notes

yelizariev:
Thank you anyway

almas:
- Because when customer doesn't selected it shouldn't decrease count then we need to check if order.get_client().
- OK
- OK

PR #27, opened 2015-10-31
Currently:

bassn:
(1) If user clicks a customer with debt, then button.set-customer-pay-full-debt is displayed
(2) If user clicks another customer without debt, then this button is still displayed.
With this change: the button.set-customer-pay-full-debt will be hidden if a customer w/o debt is clicked again

### DIFF
--- a/pos_debt_notebook/__openerp__.py
+++ b/pos_debt_notebook/__openerp__.py
@@ -1,6 +1,6 @@
 {
     'name': 'Debt notebook (technical core)',
-    'version': '1.0.2',
+    'version': '1.0.3',
     'author': 'IT-Projects LLC, Ivan Yelizariev',
     'category': 'Point Of Sale',
     'website': 'https://yelizariev.github.io',
--- a/pos_debt_notebook/__openerp__.py
+++ b/pos_debt_notebook/__openerp__.py
@@ -1,6 +1,6 @@
 {
     'name': 'Debt notebook (technical core)',
-    'version': '1.0.3',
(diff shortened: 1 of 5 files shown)


yelizariev:
@bassn thank you for contribution. I'm ready to accept. Could you please bump 1.0.3 version (in openerp file and in doc/changelog)

bassn:
@yelizariev : ok, i will do so. 
I would also remove 'account' from dependencies of module 'pos_debt_notebook', as it is already in dependencies of module 'point_of_sale'.
Is that ok, for you?

yelizariev:
> I would also remove 'account' from dependencies of module 'pos_debt_notebook', as it is already in dependencies of module 'point_of_sale'.

I set this depency explicitly to show that we use account features. So, it's a part of documentation. Probably it's better to return it.

yelizariev:
Also, you add changelog to pos_debt_notebook rather than tg_pos_debt_notebook module. Could you fix it, please

bassn:
ok, i put it back.

bassn:
both points are done

yelizariev:
@bassn the same about __openerp__.py -- use one in tg_pos_debt_notebook folder, rather than pos_debt_notebook

yelizariev:
I mean bumped version

bassn:
Now we have 

```
         1.0.2 in pos_debt_notebook

         1.0.3 in tg_pos_debt_notebook
```

Is that what you meant?

yelizariev:
Yes, thank you.

PR #33, opened 2015-12-02
Update tg_partner_firstname.py

OdooMod:
view reference country field, and can't install tg_module without this field


yelizariev:
As I remember, there is already country_id field in res.partner
So, you need to use this one and don't create new one

OdooMod:
@yelizariev exactly it's true. so the problem finally is in "tg_partner_view.xml", something like this "field name="country"". field in res.partner is "country_id", but in that view is "country".

yelizariev:
you can fix view

PR #63, opened 2016-02-21
Debt multi company 

flotho:
Hi,

I corrected some little problems.
The debt attribute of journal wasn't loaded in the POS.
I prevent the UI to make a useless back()
regards

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,2 @@
+*.pyc
+*.directory
--- a/pos_debt_notebook/models.py
+++ b/pos_debt_notebook/models.py
@@ -5,36 +5,44 @@ from openerp.osv import fields as old_fields
 import openerp.addons.decimal_precision as dp
 from openerp.tools.translate import _
 from openerp.exceptions import UserError
+import logging
 
+
+_logger = logging.getLogger(__name__)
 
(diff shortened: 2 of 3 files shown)


yelizariev:
@ufaks please review these updates

yelizariev:
@flotho we are sorry for delaying for this update and thank you for your contribution.

I have a question about this commit https://github.com/yelizariev/pos-addons/pull/63/commits/a6adfdfc3989dc14f19e902ca47f4332028f4217
as I understand, you add a possibility to have multiple debt journal in the same company.
Could you give an example when business needs multiple debt journal?

flotho:
Hi @yelizariev ,
I don't know if it's a correct way to do this but our customer have many different reason of dealing with kind of debt. 
- If the customer want to pay later an don't let any cash nor check
- If the customer let multiple checks
- If we refund him
- etc...
  Thos cas are clearly identified in the customer process and have to be easily identified, that's why we're working with multiple journal in the POS. And some of them have to give some debts etc...
  Of course account linked to those accounts are sometimes the same but we have multiple journals giving some debt.

What's your point of view?

regards

yelizariev:
Merged here https://github.com/yelizariev/pos-addons/pull/92

PR #74, opened 2016-03-04
[FIX]load quantity in pos cache

gaelTorrecillas:
closes https://github.com/yelizariev/pos-addons/issues/72


yelizariev:
As I understand key commit  here is this one
https://github.com/yelizariev/pos-addons/commit/6a9d18185811ed3569229148d0c1f42fa92aa2a8

I don't think that your solution works with pos_cache. While you will get qty_available field, it will have cached value

yelizariev:
Please, create separate branch for each update you made

gaelTorrecillas:
@yelizariev 

with old method, the pos_cache don't load qty_available, but if we push this field with the new field, so the pos cache load this field.

yelizariev:
I understand, but pos_cache load _cached_ value of qty_available, not a fresh one. Try to make stock updates and reload page

gaelTorrecillas:
We understand your position, but actually, modification works, the values are calculated every night
thank you for your comment
the commit  pos_multi_sessions :withdrawal of customer (31bfdc2c5c0dcfe190c91573f494268f35af4ed7) delete the client if he's removed in the PoS (pos_multi_sessions)

sorry I forgot to create new branch

yelizariev:
ok, I'll add this commit https://github.com/yelizariev/pos-addons/commit/6a9d18185811ed3569229148d0c1f42fa92aa2a8 soon

we will add https://github.com/yelizariev/pos-addons/commit/31bfdc2c5c0dcfe190c91573f494268f35af4ed7 to pos_multi_sessions later, because we need to make backport, update version etc

gaelTorrecillas:
ok, no problem

yelizariev:
6a9d181 in 9.0 now

PR #79, opened 2016-03-07
9.0 fix multi session note

cvinh:
In pos_multi_session, line notes are now synchronized between poses sharing the multisession


cvinh:
Bad PR because note come from pos_restaurant module

PR #102, opened 2016-03-29
[FIX] fix pos_session_custom2 not working in v9

cvinh:
I tested pos_session_custom2 and it did not work in actual v9
Here is the fix

### DIFF
--- a/pos_session_custom2/model/session.py
+++ b/pos_session_custom2/model/session.py
@@ -269,11 +269,11 @@ class sessionpos(osv.Model):
         for session in self.browse(cr,uid,ids,context=context):
             for order in session.order_ids:
                 for line in order.lines:
-                    taxes_ids = [ tax for tax in line.product_id.taxes_id if tax.company_id.id == line.order_id.company_id.id ]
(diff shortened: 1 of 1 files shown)


cvinh:
small question here : how do you print backend pdf receipt to posbox printer ?

yelizariev:
As for now, It's not possible to print backend pdf to posbox

cvinh:
Would it be good for you to move the session report in frontend ?
Then we could print it on the posbox...

yelizariev:
I don't have such plans

cvinh:
Dear Yvan, here's a tip for you : you can print pdf on a posbox printer by installing a cups raw queue in the posbox (it's readonly, you have to hack it)... Then you can print the pdf session report from any computer in the local network (install the printer driver on every computer)
We did it successfully

yelizariev:
Do you mean, that you are able to print pdf file on ticket printer?
images

cvinh:
yes, a ticket printer (epson tm-t20ii in my case) plugged on a posbox (with cups running on it)

PR #108, opened 2016-04-13
delete product_variant_multi as it relocated to OCA

navarromiguel:
Updated product_variant_multi addon to  Odoo 9 and set installable to True. It has been tested and it's apparently okay.


cvinh:
Hello
This module is maintained here : https://github.com/OCA/product-variant
I think it would be better to do your PR there
Except if @yelizariev maintains the module here

yelizariev:
@invitu 
thank you for the link.
I will remove product_variant_multi from this repository

PR #112, opened 2016-04-24
9.0 Add currency in pos session report

cvinh:
Add currency in pos session report

### DIFF
--- a/pos_session_custom2/views/report1.xml
+++ b/pos_session_custom2/views/report1.xml
@@ -53,7 +53,8 @@
 
 					<tr t-foreach="o.statement_ids" t-as="i">
 						<td><h4><span t-field="i.journal_id"/></h4></td>
-                		<td class="text-right"><h4><span t-field="i.total_entry_encoding_custom"/></h4></td>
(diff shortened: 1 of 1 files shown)


yelizariev:
@invitu 
thank you for contribution.
I can accept you pull request only without gitignore file (use can use global gitignore confguration) and without changing indent, because it bloats git blame

cvinh:
Dear @yelizariev ,
.gitignore file is used in every Odoo and OCA repository... why don't you do like them ?
I don't get your indent problem... I think I use "best practice" indentation as Odoo and OCA
Best regards

yelizariev:
@invitu the question is not about way of indentation. It's about bloating git blame https://github.com/odoo/odoo/pull/4531#issuecomment-69105136

As for gitignore I believe there is no gitignore file that suit for every developer (every IDE, file manager etc). So why don't configure global gitignore once in your machine and don't add it to every project

cvinh:
ok, I let you do these improvements

yelizariev:
Thank you

PR #116, opened 2016-05-04
[FIX]  print only those who belong to print

ilyasProgrammer:


pos_multi_session/static/src/js/pos_multi_session.js
```diff
@@ -349,6 +349,7 @@ odoo.define('pos_multi_session', function(require){
         export_as_JSON: function(){
             var data = OrderSuper.prototype.export_as_JSON.apply(this, arguments);
             data.ms_info = this.ms_info;
+            data.saved_resume = this.saved_resume;
```
yelizariev: no need to add it. saved_resume is in multiprint_resume already

PR #166, opened 2016-11-02
[ADD] added revision ID for message, function order number check

GabbasovDinar:


pos_multi_session/pos_multi_session_models.py
```diff
@@ -42,23 +42,39 @@ def on_update_message(self, message):
...
 
     @api.multi
+    def order_number_check(self, message):
+        self.ensure_one()
+        order_uid = message['data']['uid']
+        client_revision_ID = message['data']['revision_ID']
+        server_revision_ID = self.env['pos.multi_session.order'].search([('order_uid', '=', order_uid)]).revision_ID
+        if client_revision_ID is not server_revision_ID:
+            print "The client is not updated"
+            return False
```
yelizariev, line 52: I guess you need to raise an Exception here

pos_multi_session/pos_multi_session_models.py
```diff
@@ -42,23 +42,39 @@ def on_update_message(self, message):
         return res
 
     @api.multi
+    def order_number_check(self, message):
+        self.ensure_one()
+        order_uid = message['data']['uid']
+        client_revision_ID = message['data']['revision_ID']
+        server_revision_ID = self.env['pos.multi_session.order'].search([('order_uid', '=', order_uid)]).revision_ID
+        if client_revision_ID is not server_revision_ID:
+            print "The client is not updated"
```
yelizariev: remove print, use `_logger`

pos_multi_session/pos_multi_session_models.py
```diff
@@ -42,23 +42,39 @@ def on_update_message(self, message):
...
+            return False
+
+    @api.multi
     def set_order(self, message):
         self.ensure_one()
+        self.order_number_check(message)
         msg_data = message['data']
         order_uid = msg_data['uid']
         order = self.env['pos.multi_session.order'].search([('order_uid', '=', order_uid)])
```
yelizariev, line 60: it's better to call `order_number_check` here to don't search order twice

pos_multi_session/pos_multi_session_models.py
```diff
@@ -42,23 +42,39 @@ def on_update_message(self, message):
         return res
 
     @api.multi
+    def order_number_check(self, message):
```
yelizariev: rename to `check_order_revision` for example

pos_multi_session/pos_multi_session_models.py
```diff
@@ -42,23 +42,39 @@ def on_update_message(self, message):
...
         order_uid = msg_data['uid']
         order = self.env['pos.multi_session.order'].search([('order_uid', '=', order_uid)])
         if order:  # order already exists
             order.write({
                 'order': json.dumps(message),
+                'revision_ID': order.revision_ID + 1,
             })
         else:
-            order.create({
+            res = order.create({
```
yelizariev: you can use `order` here instead of `res`

pos_multi_session/pos_multi_session_models.py
```diff
@@ -78,7 +95,9 @@ def remove_order(self, message):
         self.ensure_one()
         msg_data = message['data']
         order_uid = msg_data['uid']
-        self.order_ids.search([('order_uid', '=', order_uid)]).unlink()
+        order = self.order_ids.search([('order_uid', '=', order_uid)])
+        message['data']['revision_ID'] = order.revision_ID
```
yelizariev: Probably, we don't need to send `revision_ID` here, because removing cannot be under

pos_multi_session/pos_multi_session_models.py
```diff
@@ -112,6 +131,7 @@ class PosMultiSessionOrder(models.Model):
 
     order = fields.Text('Order JSON format')
     order_uid = fields.Char()
+    revision_ID = fields.Integer(default=1, string="Last sent order number")
```
yelizariev:
```
 string="Revision", help="Number of updates received from clients"
```

pos_multi_session/static/src/js/pos_multi_session.js
```diff
@@ -216,6 +218,7 @@ openerp.pos_multi_session = function(instance){
                     line.uid = dline.uid;
                 }
                 line.ms_info = dline.ms_info || {};
+                line.revision_ID = dline.revision_ID || {};
```
yelizariev: As I understand order line doesn't have `revision_ID`.

pos_multi_session/static/src/js/pos_multi_session.js
```diff
@@ -275,11 +278,14 @@ openerp.pos_multi_session = function(instance){
             options = options || {};
             OrderSuper.prototype.initialize.apply(this, arguments);
             this.ms_info = {};
+            this.revision_ID = 1;
```
yelizariev:
```
  this.revision_ID = options.revision_ID or 1;
```

pos_multi_session/static/src/js/pos_multi_session.js
```diff
@@ -275,11 +278,14 @@ openerp.pos_multi_session = function(instance){
...
             if (!_.isEmpty(options.ms_info)){
                 this.ms_info = options.ms_info;
+                this.revision_ID = options.revision_ID;
```
yelizariev: delete it

pos_multi_session/static/src/js/pos_multi_session.js
```diff
@@ -338,8 +345,11 @@ openerp.pos_multi_session = function(instance){
                 if (error == 'offline') {
                     self.order_on_server = false;
                 }
-            }).done(function(){
+            }).done(function(id){
```
yelizariev:
```
        }).done(function(server_revision_ID){
```

pos_multi_session/static/src/js/pos_multi_session.js
```diff
@@ -349,6 +359,7 @@ openerp.pos_multi_session = function(instance){
             var self = this;
             OrderlineSuper.prototype.initialize.apply(this, arguments);
             this.ms_info = {};
+            this.revision_ID = 1;
```
yelizariev: Do we need revision_ID per each order line?

pos_multi_session/static/src/js/pos_multi_session.js
```diff
@@ -372,6 +383,7 @@ openerp.pos_multi_session = function(instance){
             var data = OrderlineSuper.prototype.export_as_JSON.apply(this, arguments);
             data.uid = this.uid;
             data.ms_info = this.ms_info;
+            data.revision_ID = this.revision_ID;
```
yelizariev: to delete

pos_multi_session/static/src/js/pos_multi_session.js
```diff
@@ -443,13 +455,15 @@ openerp.pos_multi_session = function(instance){
...
                 }
-                connection_status.resolve();
                 self.show_warning_message = true;
                 if (Array.isArray(res)) {
                     res.forEach(function(item) {
                         if (item.action == 'sync_all') self.pos.ms_on_update(item);
                     });
                 }
+                if (res.action == "update_revision_ID") {
+                    connection_status.resolve(res.revision_ID);
```
yelizariev: It has to be always resolved or rejected. You cannot just ignore it.

Don't forget, that function `send` is not used for `update_order` only and server doesn't return `update_revision_ID` always.

yelizariev (changes requested):
You broke `send` function in js and you don't really check revision_ID in server side (result of the function order_number_check is ignored)

Everything else is not bad.

pos_multi_session/static/src/js/pos_multi_session.js
GabbasovDinar: да, именно поэтому делается проверка res.action == "update_revision_ID"
ну и добавил условие else connection_status.resolve()

pos_multi_session/pos_multi_session_models.py
GabbasovDinar: наверное тут надо будет вернуть клиенту какое то значение, чтоб знать какая именно ошибка вышла

yelizariev:
Fix lint

```

======== Testing test_flake8 ========
pos_multi_session/pos_multi_session_models.py:9:1: F401 'openerp.osv.osv' imported but unused
pos_multi_session/pos_multi_session_models.py:10:1: F401 'openerp.tools.translate._' imported but unused
pos_multi_session/pos_multi_session_models.py:11:1: F401 'openerp.exceptions.Warning as UserError' imported but unused
======== Testing test_pylint ========
From https://github.com/it-projects-llc/pos-addons
 * [new branch]      8.0-pos_multi_session-2.0.0 -> origin/8.0-pos_multi_session-2.0.0
Start lint check just in modules changed
************* Module pos_multi_session
pos_multi_session/doc/changelog.rst:1: [E7901(rst-syntax-error), ]  Hyperlink target "changelog" is not referenced.
pos_multi_session/static/src/js/pos_multi_session.js:36: [W7903(javascript-lint), ] Missing semicolon.
pos_multi_session/static/src/js/pos_multi_session.js:445: [W7903(javascript-lint), ] Missing semicolon.
pos_multi_session/static/src/js/pos_multi_session.js:1: [W7908(missing-newline-extrafiles), ]  Missing newline
************* Module pos_multi_session.pos_multi_session_models
pos_multi_session/pos_multi_session_models.py:9: [W0402(deprecated-module), ] Uses of a deprecated module 'openerp.osv'
pos_multi_session/pos_multi_session_models.py:127: [E8102(invalid-commit), PosMultiSession.broadcast_message] Use of cr.commit() directly - More info https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#never-commit-the-transaction

```

pos_multi_session/external_tests/tests/test_sync.py
```diff
@@ -219,7 +219,7 @@ def test_30_slow(self):
              "code": """
                  mstest.print_order();
                  synced_order = mstest.get_order();
-                 mstest.check_inclusion(share.demo_order, synced_order);
+                 // mstest.check_inclusion(share.demo_order, synced_order);
```
yelizariev, line 223: add a comment

// this test is currently not supported: request from demo is ignored and hist state is rollbacked

pos_multi_session/static/src/js/pos_multi_session.js
```diff
@@ -455,26 +457,31 @@ openerp.pos_multi_session = function(instance){
                     self.offline_sync_all_timer = false;
                     self.send_offline_orders();
                 }
+                if (res.action == "update_revision_ID") {
+                    connection_status.resolve(res.revision_ID);
+                }
+                connection_status.resolve();
+                if (res.action == "revision_error") {
+                    var warning_message = _t('The conflict during of synchronization, repeat your operation');
```
yelizariev:
```
                var warning_message = _t('There is a conflict during synchronization, try your action again');
```

yelizariev (changes requested):
Tiny style updates are required

PR #168, opened 2016-11-09
8.0 pos multi session 

GabbasovDinar:


pos_multi_session/pos_multi_session_models.py
```diff
@@ -67,15 +68,19 @@ def set_order(self, message):
                 'revision_ID': order.revision_ID + 1,
             })
         else:
+            if self.order_ID >= sequence_number:
```
yelizariev: use more exact check, probably 

```
        if self.order_ID +1 != sequence_number:
```

pos_multi_session/pos_multi_session_models.py
```diff
@@ -89,6 +94,9 @@ def get_sync_all(self, message):
             msg['data']['revision_ID'] = order.revision_ID
             msg['action'] = 'sync_all'
             message.append(msg)
+        if not message:
+            order_ID = self.env["pos.multi_session"].search([('id', '=', self.id)]).order_ID
```
yelizariev:
```
        order_ID = self.order_ID
```

pos_multi_session/static/src/js/pos_multi_session.js
```diff
@@ -299,10 +315,21 @@ openerp.pos_multi_session = function(instance){
         set_client: function(client){
             OrderSuper.prototype.set_client.apply(this,arguments);
             this.trigger('change:sync');
+            this.update_sequence_number();
         },
         addProduct: function(){
+            var self = this;
             OrderSuper.prototype.addProduct.apply(this, arguments);
             this.trigger('change:sync');
+            this.update_sequence_number();
```
yelizariev: call this in ms_update instead

pos_multi_session/static/src/js/pos_multi_session.js
```diff
@@ -345,8 +372,15 @@ openerp.pos_multi_session = function(instance){
                 if (error == 'offline') {
                     self.order_on_server = false;
                 }
-            }).done(function(server_revision_ID){
+            }).done(function(server_revision_ID, order_ID){
```
yelizariev: use a dictionary here

```
        }).done(function(res){

  res.order_ID ....
```

yelizariev (changes requested):

yelizariev:
Fix lint errors too

PR #170, opened 2016-11-18
[FIX] synchronization error the remove order when client offline fix

GabbasovDinar:


pos_multi_session/static/src/js/pos_multi_session.js
```diff
@@ -70,9 +70,12 @@ openerp.pos_multi_session = function(instance){
                     this.get('selectedOrder').ms_replace_empty_order = true;
                     return;
                 } else if (this.ms_syncing_in_progress){
-                    if (this.get('orders').size() === 0){
+                    if (this.get('orders').size() === 0) {
                         this.add_new_order();
                     }
+                    if (this.get('orders').size() > 0){
```
yelizariev: use `else` here

pos_multi_session/static/src/js/pos_multi_session.js
```diff
@@ -507,16 +505,50 @@ openerp.pos_multi_session = function(instance){
                     self.warning(warning_message);
                     self.request_sync_all();
                 }
-
-                self.show_warning_message = true;
+                if (res.action == 'sync_order_id') {
+                    self.pos.pos_session.order_ID = res.order_ID;
+                    self.pos.pos_session.sequence_number = res.order_ID;
+                    self.destroy_remove_orders(server_orders_uid);
```
yelizariev: destroy_remove**d**_orders

yelizariev: `server_orders_uid` is always empty here

pos_multi_session/static/src/js/pos_multi_session.js
```diff
@@ -507,16 +505,50 @@ openerp.pos_multi_session = function(instance){
...
+                }
                 if (Array.isArray(res)) {
+                    // load all orders
```
yelizariev: _load_ usually means, that you download orders from servers

pos_multi_session/static/src/js/pos_multi_session.js
```diff
@@ -507,16 +505,50 @@ openerp.pos_multi_session = function(instance){
...
+                    clearInterval(self.offline_sync_all_timer);
+                    self.offline_sync_all_timer = false;
                 }
+                self.show_warning_message = true;
             });
             return connection_status;
         },
+        destroy_remove_orders: function(server_orders_uid) {
+            var self = this;
+            // find all the client orders whose order_on_server is True
```
yelizariev:
```
        // find all client orders whose order_on_server is True
```

yelizariev (changes requested):
add `else`, make style updates

yelizariev:
make rebase

PR #176, opened 2016-11-21
8.0 pos multi session. test error FIX

GabbasovDinar:


pos_multi_session/external_tests/tests/test_sync.py
```diff
@@ -125,9 +125,9 @@ def test_20_offline_update_order(self):
             # demo creates new order
             {"session": "demo",
              "code": """
-                mstest.new_order();
+                mstest.fill_order();
```
yelizariev, line 128: I cannot accept replacing new_order to fill_order. We have to figure out why it doesn't work with new_order

yelizariev:
superseded by https://github.com/it-projects-llc/pos-addons/pull/177

PR #192, opened 2016-12-06
[MIG] migrate to_pos_shared_floor to v10

cvinh:


yelizariev:
Dear @invitu 
All modules are already ported and work in v10. While updating openerp to odoo in importing makes sense, I cannot accept other updates that just fix indentation and rename files. It would lead to merging problems if we make updates in 9.0 or 8.0 branches. Also, it blows up git history.  Hope for your understanding

cvinh:
ok, I understand your PoV, if it still works with the openerp, I have no problem

PR #193, opened 2016-12-06
[MIG] migrate pos_disable_payment to v10

cvinh:


yelizariev:
see https://github.com/it-projects-llc/pos-addons/pull/192#issuecomment-265147119

PR #194, opened 2016-12-06
10.0 mig pos multisession

cvinh:


yelizariev:
see https://github.com/it-projects-llc/pos-addons/pull/192#issuecomment-265147119

PR #198, opened 2016-12-17
[FIX] function is_product_in_category should overwrite original function

cvinh:
tested in v10, the javascript stucks on pos_categ_id that is not available


yelizariev:
You are right. 
Also, this update can be applied for other branches too.

By the way, feel free to add yourself (or your company) to contributors section in README.rst file whenever you send updates

PR #199, opened 2016-12-17
[FIX] view dependance should be done on point_of_sale module instead …

cvinh:
…of product module
tested with v10


yelizariev:
Thank you

PR #200, opened 2016-12-21
[FIX] increase collecting timer to avoid "sync conflict" problem.

yelizariev:
The reason of the problem is that some update request could be delivered to
server in a wrong order. If we increase timer, we descrease probability of such
situation, because minimal time between two requests is 1 sec now


yelizariev:
@ilmir-k try to repeat the issue here

PR #201, opened 2016-12-22
[FIX] repeats access error message fix

GabbasovDinar:


yelizariev:
Are you sure, that we don't need to send request_sync_call in case of access_error?

yelizariev:
What access_error actually is?

GabbasovDinar:
@yelizariev, I think yes. When I run the synchronization if access_error then he causes sync is infinitely

yelizariev:
What access_error  is?

GabbasovDinar:
@yelizariev exception type

yelizariev:
You can check whether it was sync_all request and don't send new one and show another wanrning ("Cannot sync") instead

yelizariev:
I mean when access_error happens?

yelizariev:
what is the reason of the error?

PR #204, opened 2016-12-26
[FIX] negative debt

ufaks:


pos_debt_notebook/static/src/js/pos.js
```diff
@@ -53,6 +55,22 @@ odoo.define('pos_debt_notebook.pos', function (require) {
             }
             this.paymentlines.add(newPaymentline);
             this.select_paymentline(newPaymentline);
+        },
+        get_due: function(paymentline) {
```
yelizariev: don't modify built-in method. Create new one instead

yelizariev (changes requested):

pos_debt_notebook/static/src/js/pos.js
```diff
@@ -49,10 +51,26 @@ odoo.define('pos_debt_notebook.pos', function (require) {
             var newPaymentline = new models.Paymentline({},{order: this, cashregister: cashregister, pos: this.pos});
             if(cashregister.journal.type !== 'cash' || this.pos.config.iface_precompute_cash ||
                 cashregister.journal.debt){
-                newPaymentline.set_amount( Math.max(this.get_due(),0) );
+                newPaymentline.set_amount( this.get_due_debt());
```
yelizariev: Use ``get_due_debt`` only for debts

yelizariev (changes requested):

PR #212, opened 2017-01-29
[IMP] add ability to change a credit limit in POS

ufaks:


pos_debt_notebook/static/src/js/pos.js
```diff
@@ -79,29 +79,36 @@ odoo.define('pos_debt_notebook.pos', function (require) {
...
-
             if(isDebt && currentOrder.get_orderlines().length === 0){
                 this.gui.show_popup('error',{
                     'title': _t('Empty Order'),
                     'body': _t('There must be at least one product in your order before it can be validated. (Hint: you can use some dummy zero price product)'),
                 });
                 return;
             }
-
+            if (client.debt + debt_amount > client.credit_limit) {
```
yelizariev: Rename ``credit_limit`` to ``debt_limit``. Let term Credit be for positive value (prepayment), while debt is negative

pos_debt_notebook/models.py
```diff
@@ -56,12 +61,24 @@ def _compute_debt(self):
...
 
     def _compute_debt_type(self):
         debt_type = self.env["ir.config_parameter"].get_param("pos_debt_notebook.debt_type", default='debt')
         for partner in self:
             partner.debt_type = debt_type
 
+    @api.model
+    def create_from_ui(self, partner):
+        if partner.get('credit_limit'):
+            if not self.env.user.has_group('point_of_sale.group_pos_manager'):
```
yelizariev: check it on write\create methods instead

pos_debt_notebook/models.py
```diff
@@ -56,12 +61,24 @@ def _compute_debt(self):
...
+                raise exceptions.Warning('Only POS managers can change a credit limit value!')
+        else:
+            partner['credit_limit'] = self._default_credit_limit()  # Default functions are not called in standard way
```
yelizariev: it is not called because partner already have ``credit_limit`` attribute (equal to false)

pos_debt_notebook/models.py
yelizariev: add a method ``check_access_to_debt_limit``, call it here and if user doesn't have it remove value from the ``partner``.
Also, you can add ``check_access_to_debt_limit``, to write\create methods:

```
if credit_limit in vals and not self.check_access_to_debt_limit():
    raise ...
```

pos_debt_notebook/models.py
```diff
@@ -192,6 +209,9 @@ class PosConfiguration(models.TransientModel):
         ('credit', 'Display Credit')
     ], default='debt', string='Debt Type', help='Way to display debt value (label and sign of the amount). '
                                                 'In both cases debt will be red, credit - green')
+    credit_limit = fields.Float(
+        string='Credit Limit', digits=dp.get_precision('Account'), default=0,
+        help='Customers cannot buy products on credit over this value')
```
yelizariev: debt_limit = fields.Float(
        string='Max Debt', digits=dp.get_precision('Account'), default=0,
        help='Customer is not allowed to have Debt more than this value')

pos_debt_notebook/models.py
```diff
@@ -200,3 +220,11 @@ def set_debt_type(self):
     def get_default_debt_type(self, fields):
         debt_type = self.env["ir.config_parameter"].get_param("pos_debt_notebook.debt_type", default='debt')
         return {'debt_type': debt_type}
+
+    def set_credit_limit(self):
+        for record in self:
```
yelizariev: no need to make iteration here. ``res.config`` is always one record

yelizariev (changes requested):

pos_debt_notebook/models.py
```diff
@@ -61,22 +62,35 @@ def _default_credit_limit(self):
...
+        help='The customer is not allowed to have a debt more than this value')
 
     def _compute_debt_type(self):
         debt_type = self.env["ir.config_parameter"].get_param("pos_debt_notebook.debt_type", default='debt')
         for partner in self:
             partner.debt_type = debt_type
 
+    def check_access_to_debt_limit(self, vals):
+        debt_limit = vals.get('debt_limit')
+        if (not self.env.user.has_group('point_of_sale.group_pos_manager') and debt_limit and
```
yelizariev: It allows anybody change value to zero.

pos_debt_notebook/models.py
```diff
@@ -61,22 +62,35 @@ def _default_credit_limit(self):
...
+                float(self._default_debt_limit()) != debt_limit):
```
yelizariev: move flaot convertion to _default_debt_limit

yelizariev (changes requested):

PR #220, opened 2017-02-03
[ADD] added queue to requests

GabbasovDinar:


pos_debt_notebook/static/src/js/pos.js
```diff
@@ -14,6 +14,8 @@ odoo.define('pos_debt_notebook.pos', function (require) {
     var _super_posmodel = models.PosModel.prototype;
     models.PosModel = models.PosModel.extend({
         initialize: function (session, attributes) {
+            this.partner_list_ids = [];
```
yelizariev: reload_debts_partner_ids

yelizariev (changes requested):

PR #221, opened 2017-02-03
[FIX] pos_longpolling: fix domain to check closed session

yelizariev:
[IMP] add logs
[REF] _get_full_channel_name in python
[IMP] add_channel -- allow to pass thisArg


yelizariev:
@GabbasovDinar merged

PR #222, opened 2017-02-03
[IMP] add timeout to gather requests defore reloading

yelizariev:
[REF] handle data in a seaparate function _on_load_debts


yelizariev:
@GabbasovDinar take a look on this update too

PR #226, opened 2017-02-03
[FIX] Fixes refactoring errors

GabbasovDinar:


yelizariev:
updates seems ok for now

PR #241, opened 2017-03-02
[IMP] Add Allow to Create Order Line access rights

cvinh:
I have a case like that : 
- the screen is opened by the Manager with a generic user (ex pos1) that does not have the right to make an order (we want only real users to make an order)
- the real user (ex John Doe) has to identify himself with a PIN code or a barcode to make an order.
- When John Doe has finished his order, I comes back to the generic user (who has no PIN code)
- Next waiter that uses the pos has to identify himself too


yelizariev:
@ilmir-k review this update please

ilmir-k (approved):
:+1:

PR #248, opened 2017-03-04
[FIX] Fixes revision errors in 10.0 pos multi session

GabbasovDinar:


ilmir-k:
@GabbasovDinar you need to rebase your branch

PR #250, opened 2017-03-06
10.0 pos debt notebook improve work on network issues

yelizariev:


pos_longpolling/static/src/xml/pos_longpolling_connection.xml
```diff
@@ -1,9 +1,9 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <templates id="template" xml:space="preserve">
     <t t-extend="SynchNotificationWidget">
-        <t t-jquery=".js_msg" t-operation="after">
+        <t t-jquery=".js_synch" t-operation="append">
             <div class="js_poll_connected oe_icon oe_red">
-                <i class='fa fa-fw fa-signal'></i>
+                <i class='fa fa-fw fa-plug'></i>
```
GabbasovDinar, line 6: @yelizariev We have fa-plug icon for odoo 9 and odoo 10 only it's don't exist in odoo 8

yelizariev: ok, why can't we use fa-plug for 9 and 10 and use something else in 8?

PR #257, opened 2017-03-09
[FIX] Refactoring code and fixes connection status error

GabbasovDinar:


yelizariev:
Refactoring is good. Not sure about error fixing

yelizariev:
[ADD] Added help for timeout fields

It has to be a ``[DOC]`` tag. Check specification for tags: https://github.com/it-projects-llc/odoo-development/blob/master/docs/git/commits.rst

PR #262, opened 2017-03-21
[FIX] Fixes the error of changing the order after exiting to main floor

GabbasovDinar:


ilmir-k:
@GabbasovDinar 
Add new version

ilmir-k:
@GabbasovDinar fix lint errors

PR #263, opened 2017-03-21
[FIX] Fixes errors for pos_multi_session

GabbasovDinar:


pos_multi_session/models/pos_multi_session_models.py
```diff
@@ -152,13 +148,11 @@ class PosSession(models.Model):
...
-                orders = self.config_id.multi_session_id.order_ids.search([('state', '!=', 'deleted')])
-                for order in orders:
-                    order.state = 'deleted'
+        self.config_id.write({'multi_session_message_ID': 1})
+        active_sessions = self.env['pos.session'].search([('state', '!=', 'closed'), ('config_id.multi_session_id', '=', self.config_id.multi_session_id.id)])
+        if len(active_sessions) == 0:
+            self.config_id.multi_session_id.sudo().write({'order_ID': 0})
+            orders = self.config_id.multi_session_id.order_ids.filtered(lambda x: x.state == "draft")
+            for order in orders:
+                order.state = 'unpaid'
```
yelizariev: use single call for write

    orders.write({'state': 'unpaid'})

yelizariev (changes requested):

ilmir-k (approved):
:+1:

PR #282, opened 2017-04-20
[FIX] Reference Error: Model is not defined for pos_multi_session

GabbasovDinar:


.travis.yml
```diff
@@ -2,8 +2,7 @@ language: python
 
 python:
   - "2.7"
-  
-dist: trusty
```
yelizariev: revert it back please

GabbasovDinar: @yelizariev DONE

Issue #281, opened 2017-04-20
[9.0][10.0] problem to inherit push_order


gaelTorrecillas:
hello, 

in our inheritance of push_order, we lose the options.

https://github.com/it-projects-llc/pos-addons/blob/10.0/pos_product_available/static/src/js/pos.js#L26

fix in commit : https://github.com/MindAndGo/pos-addons/commit/3697a71b443c9dcd627c039c4caa1b6e0db030e6

yelizariev:
Hello, 
thank you for update

@ilmir-k, check and apply this commiit please

PR #287, opened 2017-04-25
[FIX][9.0] pos_product_available - don't lose the options when inheritance of push_order

ilmir-k:
[FIX][9.0] pos_product_available - don't lose the options when inheritance of push_order


yelizariev:
?

ilmir-k:
https://github.com/it-projects-llc/pos-addons/issues/281

PR #289, opened 2017-04-26
[FIX] lint errors v10

ufaks:


ilmir-k:
@ufaks fix the remaining lint errors please

PR #293, opened 2017-04-27
[ADD] a more load button in POS debt history

ufaks:


pos_debt_notebook/static/src/js/pos.js
```diff
@@ -450,6 +452,25 @@ odoo.define('pos_debt_notebook.pos', function (require) {
                     debt_history_line = debt_history_line.childNodes[1];
                     contents.appendChild(debt_history_line);
                 }
+                if (debt_history.length % this.debt_history_limit === 0) {
+                    var debt_history_load_more_html = QWeb.render('DebtHistoryLoadMore');
+                    var debt_history_load_more = document.createElement('tbody');
+                    debt_history_load_more.innerHTML = debt_history_load_more_html;
+                    debt_history_load_more = debt_history_load_more.childNodes[1];
+                    contents.appendChild(debt_history_load_more);
+                    var new_limit = $('#debt_history tr').length - 2 + this.debt_history_limit;
```
yelizariev: You cannot do it according to mvc approach. Fix that please

yelizariev (changes requested):
https://en.m.wikipedia.org/wiki/Model%E2%80%93view%E2%80%93controller

pos_debt_notebook/static/src/js/pos.js
```diff
@@ -398,6 +405,8 @@ odoo.define('pos_debt_notebook.pos', function (require) {
             this.pos.on('updateDebtHistory', function(partner_ids){
                 this.update_debt_history(partner_ids);
             }, this);
+            this.debt_history_limit = 10;
+            this.debt_history_built_lines = 0;
```
yelizariev: Can it be more clear names? Also, use a separate setting for initial limit.

E.g.

            this.debt_history_limit_initial = 10;
            this.debt_history_limit_increment = 20;
            this.debt_history_built_current = 0;

yelizariev (changes requested):

manu686:
yay 👍

PR #295, opened 2017-04-29
[FIX] [10.0] pos_multi_session_restaurant: Refresh floor screen on removed order

ata2001:
`pos_multi_session_restaurant` does not refresh the floor screen on removed order, when only one table has orders. Could fix with triggering `'change:orders-count-on-floor-screen'` on removed order.


yelizariev:
Thank you @ata2001 

@ilmir-k or @GabbasovDinar  review this please

ilmir-k (approved):
:+1:

PR #298, opened 2017-05-05
[PORT] 10.0-product_lot

ArtyomLosev:


ilmir-k:
@yelizariev check travis errors please. 
we have not lint errors now, but here is another strange ones.

PR #310, opened 2017-06-16
[DOC] print method new logo

Englishrain:
[DOC] print method new logo


ilmir-k:
@yelizariev it can be merged

PR #312, opened 2017-06-16
[DOC] pos product category discount new logo

Englishrain:
[DOC] pos product category discount new logo


ilmir-k:
@yelizariev it can be merged

Issue #321, opened 2017-07-01
[10.0-pos_multi_session_restaurant] Splitbill is not done at the right step


cvinh:
Steps to reproduce : 
- set up 2 pos belonging to the same multisession
- add items on a table
- press divide button
- select items but do not press the "Payment Button" and press the "Back Button"
Actual behavior : 
- on the other pos, the order is divided
Expected behavior : 
- on the other pos, the order should not be divided yet because it's not divided in the first pos
- on the other pos, the order has to be divided after the "Payment Button" is pressed

cvinh:
Seems to be duplicate of https://github.com/it-projects-llc/pos-addons/issues/268

PR #324, opened 2017-07-01
[FIX] Fix sync issue with splitbill

cvinh:
Fixes https://github.com/it-projects-llc/pos-addons/issues/321


GabbasovDinar (commented):
@invitu
We have one issue

Sync POS1 and POS2
In POS1 add an order with 6 items
In POS1 click on split and add 3 items to be split
In POS1 click on payment
actual behaviour:
In POS2 a new order is created which contains one item
expected behaviour:
New order must contain 3 items

cvinh:
@GabbasovDinar yes I confirm the issue...

yelizariev:
Superseded by https://github.com/it-projects-llc/pos-addons/pull/352

Issue #323, opened 2017-07-01
[10.0-pos_multi_session_restaurant] creating an empty order does not sync


cvinh:
Steps to reproduce : 
- set up 2 pos belonging to the same multisession
- click on a table to make new order without adding items
- go back to floor screen

Actual behavior : 
- on the other pos, the empty order is not created
- on the other pos, click on the same table, a new order is created
- we have 2 orders on the first pos and one order on the second pos that does the mess

Expected behavior : 
- even the empty order is created on both pos
- even the empty order is automatically deleted on the first pos (that changes odoo standard)

yelizariev:
We don't sync empty orders. The main reason was that odoo creates empty orders automatically (at least when pos_restaurant is not installed). But I understand your concerns when we work with tables. What about to change this behaviour in pos_multi_session_restaurant only?

cc @Englishrain @ilmir-k

cvinh:
hello I agree with you to change the behavior in pos_multi_session_restaurant only
I don't know what would be the scenario without tables or without pos_restaurant

cvinh:
updated https://github.com/it-projects-llc/pos-addons/pull/325

Englishrain:
Сonfirmed

cvinh:
I updated my PR to change the behavior in pos_multi_session_restaurant only as @yelizariev suggested

PR #325, opened 2017-07-02
[FIX] fix create empty order

cvinh:
Fixes https://github.com/it-projects-llc/pos-addons/issues/323


yelizariev:
Let's discuss it in #323 first

GabbasovDinar (commented):
Thank you. I will check it

Issue #332, opened 2017-07-11
[pos_debt_notebook] Partner's child debt should be configurable  [$500 awarded]


gustavovalverde:
Maybe it's a very particular case, but this could be an option.
**Company A** is using pos_debt_notebook
*Company X* is one of the clients, with a bunch of employees, that buys to Company A

I have an scenario were employees from other companies have credit on **Company A**. Those companies update that credit on a weekly basis. To have a tracking we've created each employee of Company X as a **child partner** (this way is easier to know how much Company X owes Company A, for all of it's employees consumption).

This can't be accomplished with pos_debt_notebook because the Company X childs (employees) inherit Company X accounting, and thus the accounting tab can't be modified.

I can't find a way to accomplish this using the module "as is", so I'm supposing this needs additional development. Because if an employee of Company X goes and buys something, just it's debt should get affected; as Company X could also have a different debt for it's own use.

P.d.: @yelizariev I'm willing to put a bounty on this if you see this could fit as a feature for the module.

yelizariev:
What we can do in current design of the module is to make a wizard which create multiple invoices per each employee. Then you will be able to pay them at once at menu ``Accounting >> Invoices`` (select invoices and click ``[Action] -> Register Payment``

gustavovalverde:
I'd like to start with the segregation of the `debt` field. If I assign a `parent_id `(company) to a **partner** (employee), the accounting tab is disabled in the employee, so I have no way to assign a `debt` to a partner that has a parent_id assigned.

I'll be funding this through Bountysource.

yelizariev:
I close it to be able claim the bounty. Feel free to reopen or create new issues

PR #337, opened 2017-07-21
10.0 pos debt notebook company debts

yelizariev:


yelizariev:
https://github.com/it-projects-llc/pos-addons/issues/332

ilmir-k:
I've checked it. It works fine

PR #343, opened 2017-07-31
[FIX] Fixes lint errors

GabbasovDinar:


pos_order_note/static/src/js/order_note.js
```diff
@@ -197,32 +173,60 @@ odoo.define('pos_cancel_order.order_note', function (require) {
...
-                        cancelled_exist_product.custom_notes = curr.old_custom_notes;
-                    }
+                if (j === N && k === c) {
+                    C[c++] = A[ i ];
                 }
             }
-            return res;
+            return C;
         },
+        change_custom_notes: function(current_order_custom_notes_ids, old_order_custom_notes_ids) {
```
yelizariev, line 220: use more appropriate name

yelizariev: If this function just compare two arrays you don't need so complex names for arguments

yelizariev (changes requested):

Issue #327, opened 2017-07-02
[10.0-pos_multi_session] Creating new customer in one pos makes a error in other pos


cvinh:
Steps to reproduce : 
- setup 2 poses belonging to the same multi_session
- on first pos, create a new order
- create a new customer and select it
Actual behavior : 
- on second pos, we have "ReferenceError: Can't find variable: Model"
Expected behavior : 
- no error and customer is synced in second pos

Englishrain:
confirmed

yelizariev:
@invitu it was fixed by @KolushovAlexandr 
check latest version

PR #347, opened 2017-08-09
[FIX] logs in pos ms

KolushovAlexandr:


pos_multi_session/static/src/js/pos_multi_session.js
```diff
@@ -111,23 +115,24 @@ odoo.define('pos_multi_session', function(require){
...
                     });
                 }
                 if (sync_all) {
                     this.message_ID = data.message_ID;
                     this.ms_do_update(order, data);
                 } else {
-                    if (self.message_ID + 1 != data.message_ID)
-                        self.multi_session.request_sync_all();
-                    else
+                    if (self.message_ID + 1 === data.message_ID){
```
yelizariev, line 126: You probably wanted to use ``!==``, because otherwise you change code

yelizariev: I got it. You replaced blocks. Please don't do that. Use ``!==`` instead

KolushovAlexandr: There was that lint error: Unexpected negated condition. [Error/no-negated-condition], so I fixed it according to this http://eslint.org/docs/2.0.0/rules/no-negated-condition.
Ok, I won't

KolushovAlexandr: @yelizariev What about logs? are they ok?
That issue https://github.com/it-projects-llc/pos-addons/issues/327   https://www.it-projects.info/web#id=1645&view_type=form&model=project.task also was fixed with lints.

yelizariev: logs are fine.

Ok, I didn't know about no-negated-condition.

pos_multi_session/static/src/js/pos_multi_session.js
```diff
@@ -451,10 +470,10 @@ odoo.define('pos_multi_session', function(require){
                     }
                     self.request_sync_all();
                 } else {
-                    if (!self.offline_sync_all_timer) {
-                        var warning_message = _t("No connection to the server. You can create new orders only. It is forbidden to modify existing orders.");
-                        self.warning(warning_message);
-                        self.start_offline_sync_timer();
+                    self.no_connection_warning();
+                    self.start_offline_sync_timer();
```
yelizariev: Have you removed test ``if (!self.offline_sync_all_timer) {``? Why?

yelizariev:
WIP

PR #353, opened 2017-08-17
[FIX] Canceled Reason for orderline and order, use POS only

GabbasovDinar:
pos_order_cancel
===============

* [ADD] Reason of cancel for orderline
* [FIX] Reason of cancel for order after disconnect with server
* [ADD] Auto canceling all products after cancel Order
* [ADD] All canceled orders and ordelines in Backend
* [REM] pos_restaurant from the dependence
* [TODO] update documentation

 pos_order_cancel_restaurant
========================

* [ADD] printing on kitchen printer the reasons of canceled order and orderline 
* [TODO] update documentation


pos_order_cancel/security/ir.model.access.csv
```diff
@@ -1,2 +1,3 @@
 id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
 access_pos_cancelled_reason,access_pos_cancelled_reason,model_pos_cancelled_reason,point_of_sale.group_pos_user,1,1,1,1
+access_pos_order_line_canceled,access_pos_order_line_canceled,model_pos_order_line_canceled,point_of_sale.group_pos_user,1,1,1,1
```
yelizariev: Allow only manager to remove that model records. Or noboby

pos_order_cancel/models/models.py
```diff
@@ -28,150 +32,23 @@ def _compute_state(self):
...
-        all_lines = []
-        for group_key, group_data in grouped_data.iteritems():
-            for value in group_data:
-                all_lines.append((0, 0, value),)
-        if move:  # In case no order was changed
-            move.sudo().write({'line_ids': all_lines})
-            move.sudo().post()
-        return True
+        canceled_order = self.filtered(lambda o: (not o.account_move or o.state == 'paid') and o.is_cancelled is True)
+        return super(PosOrder, self-canceled_order)._create_account_move_line(session, move)
```
yelizariev: Why not to filter uncacneled orders directly
`uncanceled_order = self.filtered(lambda o: o.account_move and o.state == 'paid'  or not o.is_cancelled)
`

pos_order_cancel_restaurant/__manifest__.py
```diff
@@ -0,0 +1,36 @@
...
+
+    "author": "IT-Projects LLC, Dinar Gabbasov",
+    "support": "apps@it-projects.info",
+    "website": "https://twitter.com/gabbasov_dinar",
+    "license": "OPL-1",
+    "price": 79.00,
+    "currency": "EUR",
+
+    "depends": [
+        "pos_order_cancel",
```
yelizariev, line 18: it shall depend on pos_restaurant, I guess

Issue #330, opened 2017-07-11
Multi section error [$75 awarded]


360org:
Hello i have error with multi section error plz help i can pay for that. plz fixed
Traceback (most recent call last):
  File "/usr/lib/python2.7/dist-packages/odoo/http.py", line 640, in _handle_exception
    return super(JsonRequest, self)._handle_exception(exception)
  File "/usr/lib/python2.7/dist-packages/odoo/http.py", line 677, in dispatch
    result = self._call_function(**self.params)
  File "/usr/lib/python2.7/dist-packages/odoo/http.py", line 333, in _call_function
    return checked_call(self.db, *args, **kwargs)
  File "/usr/lib/python2.7/dist-packages/odoo/service/model.py", line 101, in wrapper
    return f(dbname, *args, **kwargs)
  File "/usr/lib/python2.7/dist-packages/odoo/http.py", line 326, in checked_call
    result = self.endpoint(*a, **kw)
  File "/usr/lib/python2.7/dist-packages/odoo/http.py", line 935, in __call__
    return self.method(*args, **kw)
  File "/usr/lib/python2.7/dist-packages/odoo/http.py", line 506, in response_wrap
    response = f(*args, **kw)
  File "/home/odoo/en-addons/web/controllers/main.py", line 862, in call_kw
    return self._call_kw(model, method, args, kwargs)
  File "/home/odoo/en-addons/web/controllers/main.py", line 854, in _call_kw
    return call_kw(request.env[model], method, args, kwargs)
  File "/usr/lib/python2.7/dist-packages/odoo/api.py", line 679, in call_kw
    return call_kw_model(method, model, args, kwargs)
  File "/usr/lib/python2.7/dist-packages/odoo/api.py", line 664, in call_kw_model
    result = method(recs, *args, **kwargs)
  File "/home/odoo/en-addons/point_of_sale/models/pos_order.py", line 510, in create_from_ui
    pos_order = self._process_order(order)
  File "/home/odoo/pos-addons/pos_order_cancel/models/models.py", line 32, in _process_order
    order = super(PosOrder, self)._process_order(pos_order)
  File "/home/odoo/en-addons/point_of_sale/models/pos_order.py", line 117, in _process_order
    if pos_session.sequence_number <= pos_order['sequence_number']:
KeyError: 'sequence_number'

multiwebinc:
[...]

PR #355, opened 2017-08-17
[FIX] Check for correct response type to prevent undefined sequence_n…

pedroslopez:
…umber

### DIFF
--- a/pos_multi_session/static/src/js/pos_multi_session.js
+++ b/pos_multi_session/static/src/js/pos_multi_session.js
@@ -378,7 +378,7 @@ odoo.define('pos_multi_session', function(require){
                 var data = self.export_as_JSON();
                 return self.pos.multi_session.update(data).done(function(res){
                     self.order_on_server = true;
-                    if (res) {
+                    if (res && res.action=="update_revision_ID") {
                         var server_revision_ID = res.revision_ID;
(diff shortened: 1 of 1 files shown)


yelizariev:
@pedroslopez I think, that we already fixed the problem #330 with this commit: https://github.com/it-projects-llc/pos-addons/commit/a3a8685143a0e0cdc107a2a57aeedf5899460053 which makes similar thing: prevents to set empty sequence_number value

@GabbasovDinar Am I right? Shall we merge this PR?

pedroslopez:
@yelizariev I commented on the original issue (#330), but I'll quote it here just because it's related

>  This does fix the issue when it occurs when removing an order, but not when a revision_error is thrown on set_order

yelizariev:
@GabbasovDinar can you confirm [that](https://github.com/it-projects-llc/pos-addons/pull/355#issuecomment-323247494)? Can there be other responses that need to be handled same way as update_revision_ID?

GabbasovDinar:
@yelizariev , @pedroslopez Yes, I confirm this. The do_ms_update function did not take check when the server function returned a "revision error" and this fixes it.

Thanks, good work!

pedroslopez:
Great to hear!

Can this be back ported to the 9.0 branch? The same issue is present.

Issue #326, opened 2017-07-02
[10.0-pos_multi_session_restaurant] adding a note on order line do not render green


cvinh:
Steps to reproduce : 
- set up 2 pos in same multisession
- create an order and add an item that has to be printed at kitchen
- press the order submit button
- add a note on the order line

Actual behavior : 
- the line and the order submit button don't render green

Expected behavior : 
- the line and the order submit button should render to green

Englishrain:
Сonfirmed

yelizariev:
We are about to fix it. 
Note, that we will not render line green, because it's built-it behaviour after updating Notes for order lines

PR #360, opened 2017-08-21
10.0 pos ms note

KolushovAlexandr:
[FIX] #326


yelizariev:
We need update redefined update_summary instead

PR #368, opened 2017-08-25
[FIX] renamed odoo to openerp for access early versions of posBox

GabbasovDinar:


yelizariev:
Let's wait for instruction on posbox configuration

PR #374, opened 2017-08-29
[FIX] save all cancelled lines

GabbasovDinar:


yelizariev:
pos_order_cancel is checked

yelizariev:
- [ ] pos_order_cancel_restaurant - make unstallable before merging

pos_order_cancel_restaurant/__manifest__.py
```diff
@@ -33,5 +33,5 @@
     "post_init_hook": None,
 
     "auto_install": False,
-    "installable": False,
+    "installable": True,
```
yelizariev: Need to revert back before merging

yelizariev:
Message for squashed commit:

```
set of updates in pos_order_cancel and pos_order_cancel_restaurant

[FIX] 
* save line is not removed, but qty is descreased
* error message when reason is not specified
* different bugs with kitchen receipts printing
** reason printing depended on config kitchen_canceled_only
[ADD]
* new menu to review, analyse Cancelcations
* show Subtotal and taxes at canceled lines
[REF]
* add build_line_resume method in pos_restaurant_base
```

PR #378, opened 2017-08-31
[FIX] Bad Validation on Multisession Send method

chrisdilbert:
When the send_it method fails the validation tries to compare the error message with wrong string, in the fix instead to compare with an string we search if the string “XmlHttpRequestError” is present on the error message.


pos_multi_session/static/src/js/pos_multi_session.js
```diff
@@ -101,7 +101,7 @@ odoo.define('pos_multi_session', function(require){
             return PosModelSuper.prototype.on_removed_order.apply(this, arguments);
         },
         ms_on_update: function(message, sync_all){
-            this.ms_syncing_in_progress = true; // don't broadcast updates made from this message
+            this.ms_syncing_in_progress = true;
```
yelizariev, line 104: Is there any reason why have you removed the comment?

GabbasovDinar (commented):
@yelizariev This method is works

yelizariev:
It seems that #467 already have needed updates

KolushovAlexandr:
This case is already included in the further update, but in a different form

PR #384, opened 2017-09-06
[FIX] styles for buttons in Cancellation Reason popup

GabbasovDinar:


pos_order_cancel/static/src/js/widgets.js
```diff
@@ -55,7 +55,7 @@ odoo.define('pos_order_cancel.widgets', function (require) {
             // type of object which is removed (product or order)
             this.gui.show_popup('confirm-cancellation',{
                 'title': _t(title + 'Cancellation Reason'),
-                'reasons': self.pos.cancelled_reason.slice(0,8),
+                'reasons': self.pos.cancelled_reason.slice(0, 8),
```
yelizariev: I should be 9 rather than 8 I guess

yelizariev: Or pass all reasons and let your widget decide how much to show

PR #387, opened 2017-09-07
10.0 pos ms

KolushovAlexandr:


pos_multi_session_restaurant/static/src/js/pos_multi_session_restaurant.js
```diff
@@ -65,6 +65,9 @@ odoo.define('pos_multi_session_restaurant', function(require){
                     data.transfer = false;
                     return;
                 } else {
+                    if (!this.pos.config.multi_session_id){
+                        return;
+                    }
                     this.send({action: 'remove_order', data: data});
```
yelizariev: call super instead of duplicating code

pos_multi_session/views.xml
```diff
@@ -16,7 +16,8 @@
                 <xpath expr="//group[@name='features']" position="after">
                     <group string="Multi-session" name="multi_session">
                         <group>
-                            <field name="multi_session_id"/>
+                            <field name="current_session_state" invisible="1"/>
+                            <field name="multi_session_id" attrs="{'readonly':[('current_session_state', '=', 'opened')]}"/>
```
yelizariev: Add a text note : "You need to close current session to set Multi-session parameter". Show it only:
* in edit mode
* AND if current_session_state==opened

pos_multi_session/static/src/js/pos_multi_session.js
```diff
@@ -467,10 +471,16 @@ odoo.define('pos_multi_session', function(require){
             });
         },
         request_sync_all: function(){
+            if (!this.pos.config.multi_session_id){
```
yelizariev: prevent to call this function instead of checking MS here

pos_multi_session/static/src/js/pos_multi_session.js
```diff
@@ -467,10 +471,16 @@ odoo.define('pos_multi_session', function(require){
...
+                return;
+            }
             var data = {};
             return this.send({'action': 'sync_all', data: data});
         },
         remove_order: function(data){
+            if (!this.pos.config.multi_session_id){
```
yelizariev: this function is already not called when MS is false

yelizariev: this function is already not called when MS is false

pos_multi_session_restaurant/static/src/js/pos_multi_session_restaurant.js
```diff
@@ -65,6 +65,9 @@ odoo.define('pos_multi_session_restaurant', function(require){
                     data.transfer = false;
                     return;
                 } else {
+                    if (!this.pos.config.multi_session_id){
+                        return;
+                    }
```
yelizariev: this function is already not called when MS is false

yelizariev (changes requested):

pos_multi_session_restaurant/static/src/js/pos_multi_session_restaurant.js
KolushovAlexandr: @yelizariev  These lines prevent the error:
TypeError: multi_session_update() takes exactly 3 arguments (2 given)
it is called.

pos_multi_session/static/src/js/pos_multi_session.js
KolushovAlexandr: @yelizariev  TypeError: multi_session_update() takes exactly 3 arguments (2 given) - without it

pos_multi_session_restaurant/static/src/js/pos_multi_session_restaurant.js
KolushovAlexandr: on the moment of a call of the super and a line after, pos.config haven't been uploaded yet and equals to null

pos_multi_session/static/src/js/pos_multi_session.js
KolushovAlexandr: moved it to ms_check()

pos_multi_session/static/src/js/pos_multi_session.js
```diff
@@ -49,22 +49,27 @@ odoo.define('pos_multi_session', function(require){
...
-                        return false;
                     }
-                }
-                order.ms_remove_order();
+                    order.ms_remove_order();
+                });
+                self.multi_session = new exports.MultiSession(self);
+                var channel_name = "pos.multi_session";
+                var callback = self.ms_on_update;
+                self.add_channel(channel_name, callback, self);
```
yelizariev, line 71: ok

pos_multi_session_restaurant/static/src/js/pos_multi_session_restaurant.js
```diff
@@ -60,14 +60,19 @@ odoo.define('pos_multi_session_restaurant', function(require){
...
-                } else {
-                    this.send({action: 'remove_order', data: data});
                 }
-            };
+                this.multi_session.remove_order = function(data) {
+                    if (data.transfer) {
+                        data.transfer = false;
+                        return;
+                    } else {
+                        this.send({action: 'remove_order', data: data});
```
yelizariev: Call super please

yelizariev (changes requested):

PR #389, opened 2017-09-08
10.0 pos order cancel restaurant doc 2

yelizariev:


360org:
this is not public yet? i cant see in the app store

yelizariev:
@W360S I have just published it: https://www.odoo.com/apps/modules/10.0/pos_order_cancel_restaurant/

PR #393, opened 2017-09-14
[FIX] order_ID is undefined

GabbasovDinar:
This PR solves a synchronization conflict, after create a second order in POS


pos_multi_session/static/src/js/pos_multi_session.js
```diff
@@ -533,6 +533,9 @@ odoo.define('pos_multi_session', function(require){
                     self.request_sync_all();
                 }
             }).done(function(res){
+                if (!self.pos.message_ID) {
+                    self.pos.message_ID = 1;
```
yelizariev: Set original value at PosModel::initialize insead

yelizariev (changes requested):

ilmir-k:
@yelizariev it works

Issue #367, opened 2017-08-25
[10.0 - POS Debt Notebook]: Manual updates won't reset credit_balance on partner [$300 awarded]


gustavovalverde:
When using Manual Updates, I'd expect that the `credit_balance` used in the partner is also set to the same amount of the last update, but it's not, it just adds up every Manual Update done in the past to the last one.

yelizariev:
Yes, I made it to work this way. Manual update adds value instead of setting new value. What we can do here to try to improve UX is to add new field "New Balance" and compute "Balance Update" depending on "New Balance" and  "current credit_balance"

gustavovalverde:
Agree

yelizariev:
Clarification for one who will do it: 
"Balance Update" must NOT be changed (recomputed) if state is  "confirmed".

ghost:
@yelizariev Working on it: https://github.com/it-projects-llc/pos-addons/pull/396 Please review, when you'll have a time. Thank you!

PR #396, opened 2017-09-16
[ADD] New Balance field

ghost:
Closes https://github.com/it-projects-llc/pos-addons/issues/367
I'm interested to work on this, but totally new to the project. Please let me know what to correct.
Thank you!


yelizariev:
@gnidorah I was not able to check news and didn't see your PR.
It seems to be fine. Could you add some tests to increase coverage?

You can find basic information here: https://odoo-development.readthedocs.io/en/latest/dev/tests/python.html
It shall cover new function ``update_balance`` at least

yelizariev:
Here is my review

- [x] add new state "draft", because it's not a good idea to ask user to switch to "Canceled" state to apply updates
- [x] we shall not show "New Balance" field if it was not used. I.e. if user used "Balance Update" directy. Probably, we need to add new selection field, say **update_type** with values "Balance Update" and "New balance". With "New balance"  type  the field "Balance Update"  has to be readonly at any state
- [x] statusbar currently is not editable, so I can't update status
- the main purpose of #367 request is be able import csv file with new balance values, so:

  * [ ] we need a way to confirm records (i.e. change state to "Confirmed) at batch (open the menu, select records, click Action --> new button has to be appeared) -- *recommended, but not required features in terms of #367*
  * [x] update_balance is not called, i.e. **balance** field is not updated after import. Possible solution: call update_balance in ``create`` and ``write`` methods.

yelizariev:
@gustavovalverde  you are invited to confirm my review or add new comments

gustavovalverde:
Hi @gnidorah will you continue working on this based on @yelizariev comments? This is an urgent matter for me, as I'm doing this manually every day.

@yelizariev I can confirm your review. 

**Additional Comments**: After this is ready I'll make an adaptation using XMLRPC, to automate the process, as there's not supposed to be user interference. I'm telling you this so you can consider improvements were a user interaction is not _completely_ required.

P.S.: Also consider the installments will use this approach, just in case there's something additional that must be considered for that purpose.

yelizariev:
@gustavovalverde ok, I marked todo *confirming records in batch* as not required

ghost:
@yelizariev @gustavovalverde Thanks, yes doing it. Please let me know what else is wrong. [ test and batch confirm are not yet there ]

gustavovalverde:
Hi @gnidorah, is it ready with the last commit?

ghost:
Hi @gustavovalverde. Only test if left. And possible change requests from @yelizariev

pos_debt_notebook/models.py
```diff
@@ -357,7 +357,45 @@ class PosCreditUpdate(models.Model):
...
+        state = vals.get('state') or self.state
+        update_type = vals.get('update_type') or self.update_type
+        if (state == 'cancel' and update_type == 'new_balance'):
+            entries = self.search([('partner_id', '=', partner_id), ('state', '=', 'cancel'), ('update_type', '=', 'new_balance'), ('id', '<', self.id)])
+            if entries:
+                credit_balance = self.get_credit_balance(self.balance, self.new_balance)
+            else:
+                credit_balance = self.partner_id.browse(partner_id).credit_balance
+            vals['balance'] = self.get_balance(credit_balance, new_balance)
+        return super(PosCreditUpdate, self).write(vals)
```
yelizariev: Could ``create`` and ``write`` share the same code in a new common method? It's better than having same code in two places.

pos_debt_notebook/models.py
```diff
@@ -357,7 +357,45 @@ class PosCreditUpdate(models.Model):
...
+        partner_id = vals.get('partner_id') or self.partner_id.id
+        new_balance = vals.get('new_balance') or self.new_balance
+        state = vals.get('state') or self.state
+        update_type = vals.get('update_type') or self.update_type
+        if (state == 'cancel' and update_type == 'new_balance'):
+            entries = self.search([('partner_id', '=', partner_id), ('state', '=', 'cancel'), ('update_type', '=', 'new_balance'), ('id', '<', self.id)])
+            if entries:
+                credit_balance = self.get_credit_balance(self.balance, self.new_balance)
+            else:
+                credit_balance = self.partner_id.browse(partner_id).credit_balance
```
yelizariev: You should always use computed field ``credit_balance`` because it depends on records of other models and not only on ``'pos.credit.update'`` records.

yelizariev (changes requested):

yelizariev:
@gnidorah I am impressed your ability to learn new stuff. You was able to change odoo js framework which we prefer to consider as *black box* and use framework via python+xml code.

While I respect your affords in js work, it's difficult to accept it, because changing js framework may have side effects and difficult to maintain. Framework already allows to do everything without making changes in the framework code.

Also, consider to add commits instead of overriding them, because it helps to check new changes.

yelizariev:
@gnidorah Could you finish updates by Monday? Because @gustavovalverde need solution urgently and cannot wait more? I'd do it myself if there will be not updates or response by Monday.

ghost:
@yelizariev Big sorry, was busy with life tasks :( Will commit required changes today.

pos_debt_notebook/models.py
ghost: @yelizariev Please clarify. `credit_balance` is a computed field here. What exactly is meant by "always"? If you mean that in calculations I should rely only on `credit_balance` and not on `balance` values of existing records, than how it fits the ""Balance Update" must NOT be changed (recomputed) if state is "confirmed"." requirement? Only balance of confirmed records is added to `credit_balance`.

yelizariev: I meant that you must always use this code:
```
credit_balance = self.partner_id.browse(partner_id).credit_balance
```
without checking ``if entries``

yelizariev:
Please check following comments:

- [x] issue with ``if entries``
- [x] deduplicate common code in ``write`` and ``create``

ghost:
@yelizariev Done, thanks!

yelizariev:
It looks good from code-side. 

@gustavovalverde could you check it on runbot (or locally) ?

@ilmir-k please review too

pos_debt_notebook/models.py
```diff
@@ -357,7 +357,45 @@ class PosCreditUpdate(models.Model):
...
+        credit_balance = self.partner_id.browse(partner_id).credit_balance
+        vals['balance'] = self.get_balance(credit_balance, new_balance)
+
+    @api.model
+    def create(self, vals):
+        partner_id = vals.get('partner_id')
+        new_balance = vals.get('new_balance')
+        state = vals.get('state') or 'undefined'
+        update_type = vals.get('update_type') or 'undefined'
+        if ('partner_id' and 'new_balance' in vals and state == 'cancel' and update_type == 'new_balance'):
```
yelizariev: 'partner_id' is always true. You probably meant 

        if (partner_id and 'new_balance' in vals and state == 'cancel' and update_type == 'new_balance'):

pos_debt_notebook/models.py
```diff
@@ -357,7 +357,45 @@ class PosCreditUpdate(models.Model):
...
+            self.update_balance(vals, partner_id, new_balance)
+        return super(PosCreditUpdate, self).create(vals)
+
+    @api.multi
+    def write(self, vals):
+        partner_id = vals.get('partner_id') if vals.get('partner_id') is not False else self.partner_id.id
+        new_balance = vals.get('new_balance') if vals.get('new_balance') is not False else self.new_balance
+        state = vals.get('state') if vals.get('state') is not False else self.state
```
yelizariev: You can simplify it:

        state = vals.get('state') or self.state

pos_debt_notebook/models.py
```diff
@@ -357,7 +357,45 @@ class PosCreditUpdate(models.Model):
...
+
+    def update_balance(self, vals, partner_id, new_balance):
+        credit_balance = self.partner_id.browse(partner_id).credit_balance
+        vals['balance'] = self.get_balance(credit_balance, new_balance)
+
+    @api.model
+    def create(self, vals):
+        partner_id = vals.get('partner_id')
+        new_balance = vals.get('new_balance')
+        state = vals.get('state') or 'undefined'
```
yelizariev: No need for 'undefined'

    state = vals.get('state')

pos_debt_notebook/models.py
```diff
@@ -357,7 +357,45 @@ class PosCreditUpdate(models.Model):
...
+            self.update_balance(vals, partner_id, new_balance)
+        return super(PosCreditUpdate, self).create(vals)
+
+    @api.multi
+    def write(self, vals):
+        partner_id = vals.get('partner_id') if vals.get('partner_id') is not False else self.partner_id.id
+        new_balance = vals.get('new_balance') if vals.get('new_balance') is not False else self.new_balance
+        state = vals.get('state') if vals.get('state') is not False else self.state
+        update_type = vals.get('update_type') if vals.get('update_type') is not False else self.update_type
+        if (state == 'cancel' and update_type == 'new_balance'):
```
yelizariev: if ('state' != 'confirm' ...)

yelizariev (changes requested):
Create and write still have common code which can be deduplicated. 

You can see why it's necessary to do it -- updates I asked in this review you have to change twice in two different places

yelizariev:
I checked again and there is my review above. @gnidorah please check it

pos_debt_notebook/models.py
yelizariev: or even  ``if (state == 'draft' ...)`` -- we can keep canceled records untouched

pos_debt_notebook/models.py
ghost: @yelizariev This doesn't work in case I've edited an existing record and changed e.g. only `new_balance` value. Then `vals.get('state')` is `None`. 
`is not False` is here by mistake, should be `is not None`.
Adjusted code for it.

ghost:
@yelizariev Changed, thank you!

pos_debt_notebook/models.py
```diff
@@ -357,7 +357,44 @@ class PosCreditUpdate(models.Model):
...
+    ], default='draft', required=True)
+    update_type = fields.Selection([('balance_update', 'Balance Update'), ('new_balance', 'New Balance')], default='balance_update', required=True)
+
+    def get_balance(self_, balance, new_balance):
+        return -balance + new_balance
+
+    def update_balance(self, vals):
+        partner_id = vals.get('partner_id') if vals.get('partner_id') is not None else self.partner_id.id
+        new_balance = vals.get('new_balance') if vals.get('new_balance') is not None else self.new_balance
+        state = vals.get('state') if vals.get('state') is not None else self.state
```
yelizariev: use ``or draft``, because when record is created via import tool state is False here and balance is not updated

        state = vals.get('state') if vals.get('state') is not None else self.state or 'draft'

yelizariev (changes requested):
Importing doesn't work well yet

yelizariev:
I tried to make import with "New balance" and ``Balance Update`` was not computed

ghost:
@yelizariev Done, but strangely I can't reproduce this. State is a required field, so if it's not provided - there is an error "Missing required value for the field 'State' (state) at row 2" in import tool. And if it's set to draft updating works...

yelizariev:
@gnidorah you made a great job. Contact me for further cooperation, if you'd like:  yelizariev@it-projects.info

ghost:
@yelizariev Sure, thank you much!

pos_debt_notebook/views/pos_credit_update.xml
```diff
@@ -71,6 +76,20 @@
             <field name="context">{}</field>
         </record>
 
+        <record id="server_action_pos_credit_update" model="ir.actions.server">
+            <field name="name">Confirm</field>
+            <field name="model_id" ref="point_of_sale.menu_point_config_product"/>
```
yelizariev, line 81: What is it? It should be reference to model, while here is reference to menu

PR #400, opened 2017-09-22
[IMP] repleace icon

GabbasovDinar:


yelizariev:
Commit is not verfied)

yelizariev:
Fix is not correct tag

PR #423, opened 2017-10-01
[IMP] recovery of product qty after press 'cancel' button

GabbasovDinar:
recovery of product qty after press 'cancel' button in popup 'cancellation reason'


gustavovalverde:
I don't know if this is also fixing that when you Remove an Order that has notes and some were sent, with the minus (-) sign, it warns you that there was an issue syncing, but no error on the console. Weird.

ilmir-k (approved):
Approved :+1: 
test is needed only

ilmir-k:
@gustavovalverde 
Unfortunately, we were not able to reproduce syncing issue. Please try the latest updates of pos_order_cancel. Probably, it can resolve it.

PR #436, opened 2017-10-11
[10.0][ADD] pos_disable_payment: new option 'Allow refunds'

ilmir-k:
[ADD] pos_disable_payment: new option 'Allow refunds'

### DIFF
--- a/pos_disable_payment/static/src/js/tour.js
+++ b/pos_disable_payment/static/src/js/tour.js
@@ -47,6 +47,10 @@ odoo.define('pos_disable_payment.tour', function(require) {
         content: _t("<p>Click to start the point of sale interface. It <b>runs on tablets</b>, laptops, or industrial hardware.</p><p>Once the session launched, the system continues to run without an internet connection.</p>"),
         position: "bottom"
     }, {
+        content: "Switch to table or make dummy action",
+        trigger: '.table:not(.oe_invisible .neworder-button), .order-button.selected',
+        position: "bottom"
+    }, {
         trigger: '.product-list .product',
         content: _t("Add the first product"),
         position: "top"
--- a/pos_disable_payment/static/src/js/tour.js
+++ b/pos_disable_payment/static/src/js/tour.js
@@ -7,9 +7,10 @@ odoo.define('pos_disable_payment.tour', function(require) {
     var _t = core._t;
 
     tour.register('pos_disable_payment_tour', {
+        test: true,
         url: "/web",
     }, [{
-        trigger: '.o_app[data-menu-xmlid="base.menu_administration"], .oe_menu_toggler[data-menu-xmlid="base.menu_administration"]',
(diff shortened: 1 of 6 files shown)


ilmir-k:
@yelizariev review it please

yelizariev:
Basically, you need to add test for new feature:

* disable refunds
* load POS
* check that Refund button is not visible

yelizariev:
Tests don't work.
Also, there there is not check that Refund button is not visible

yelizariev:
You can update settings via python, no need tour for this

yelizariev:
.numpad-minus[style="visibility: hidden;"]

pos_disable_payment/static/src/js/tour.js
```diff
@@ -45,6 +46,16 @@ odoo.define('pos_disable_payment.tour', function(require) {
         trigger: ".o_pos_kanban button.oe_kanban_action_button",
         content: _t("<p>Click to start the point of sale interface. It <b>runs on tablets</b>, laptops, or industrial hardware.</p><p>Once the session launched, the system continues to run without an internet connection.</p>"),
         position: "bottom"
-    }]);
+    }, {
+        trigger: '.product-list .product',
```
yelizariev: This step cannot be done when pos_restaurant is installed. You need to add extra step which works both with restaurant and without one, e.g.
```

{
        content: "Swicth to table or make dummy action",
        trigger: '.table, .neworder-button:not(.oe_invisible .neworder-button)',
    }


```

yelizariev (changes requested):
Tests are not passed in travis

yelizariev:
Fix lints
don't override commits and add new ones intead

yelizariev:
tests are not passed

pos_disable_payment/static/src/js/tour.js
```diff
@@ -17,11 +47,19 @@ odoo.define('pos_disable_payment.tour', function(require) {
...
+        trigger: '.table:not(.oe_invisible .neworder-button), .order-button.selected',
+        position: "bottom"
+    }, {
+        trigger: '.product-list .product',
+        content: _t("Add the first product"),
+        position: "top"
+    }, {
+        content: _t("Hidden"),
+        trigger: '.pads .numpad-minus[style="visibility: hidden;"]',
+        position: "bottom"
```
yelizariev:
```
       content: _t("Hidden"),
        trigger: '.pads',
       run: function(){
// check that the button is hidden
}
```

yelizariev (changes requested):

yelizariev (changes requested):
remove ``_t`` from test tour -- we don't want to see those phrases in translations files

ilmir-k:
it seems that travis has broken

yelizariev:
2017-10-31 09:50:10,462 5497 ERROR openerp_test odoo.tests.common: phantomjs: error The button is not hidden

yelizariev (approved):

PR #470, opened 2017-11-02
[PORT] pos_restaurant_base from 10.0 to 11.0

GabbasovDinar:


yelizariev (approved):
code updates (without running app) is checked

pos_printer_network/static/src/js/network_printer.js
```diff
@@ -11,7 +11,7 @@ odoo.define('pos_restaurant.network_printer', function (require) {
     var chrome = require('point_of_sale.chrome');
     var PopupWidget = require('point_of_sale.popups');
     var QWeb = core.qweb;
-    var mixins = core.mixins;
+    var mixins = requere("web.mixins");
```
yelizariev: requere :grimacing:

yelizariev (changes requested):

pos_printer_network/static/src/js/network_printer.js
GabbasovDinar: @yelizariev ooops

PR #472, opened 2017-11-02
[IMP] remove old code

GabbasovDinar:


yelizariev:
PR to 9.0 is enough #471

PR #483, opened 2017-11-09
[PORT] porting from 10.0 to 11.0

GabbasovDinar:


yelizariev (approved):
code is reviewed

ilmir-k:
@GabbasovDinar add `test: true` into https://github.com/it-projects-llc/pos-addons/blob/11.0/pos_order_cancel/static/src/js/tour.js

see for example https://github.com/it-projects-llc/pos-addons/blob/11.0/pos_disable_payment/static/src/js/tour.js

yelizariev:
- [x] don't use ``_t`` for tests only tour 

                     content: _t("<p>Click <b>table</b></p>"),

yelizariev (changes requested):
don't use _t for tests only tour

yelizariev (approved):

ilmir-k (approved):
It can be merged

PR #467, opened 2017-10-31
10.0 merge pos multi session sync

yelizariev:

PR #491, opened 2017-11-16
[11.0][PORT] pos_longpolling

GabbasovDinar:


yelizariev:
We will work with pos_longpolling after https://github.com/it-projects-llc/pos-addons/pull/467

PR #497, opened 2017-11-21
[DOC] index.html: missing tag

ilmir-k:
[DOC] index.html: missing tag


ilmir-k:
@yelizariev it should be merged urgently, because we have empty description at app store

Issue #518, opened 2017-12-05
10.0 - Error in pos_multi_session_restaurant


thomaspaulb:
The error that comes is:

```
Uncaught TypeError: Cannot read property 'id' of null
http://irodion.1systeem.nl/pos_multi_session_restaurant/static/src/js/pos_multi_session_restaurant.js:130
Traceback:
TypeError: Cannot read property 'id' of null
    at child.ms_on_update (http://irodion.1systeem.nl/pos_multi_session_restaurant/static/src/js/pos_multi_session_restaurant.js:130:38)
    at http://irodion.1systeem.nl/pos_multi_session/static/src/js/pos_multi_session.js:568:34
    at Array.forEach (<anonymous>)
    at Object.<anonymous> 
(http://irodion.1systeem.nl/pos_multi_session/static/src/js/pos_multi_session.js:567:32)
    at fire (http://irodion.1systeem.nl/web/static/lib/jquery/jquery.js:3119:58)
    at Object.fireWith [as resolveWith] (http://irodion.1systeem.nl/web/static/lib/jquery/jquery.js:3231:49)
    at Object.deferred.(anonymous function) 
 (http://irodion.1systeem.nl/web/static/lib/jquery/jquery.js:3321:62)
    at fire (http://irodion.1systeem.nl/web/static/lib/jquery/jquery.js:3119:58)
    at Object.fireWith [as resolveWith] (http://irodion.1systeem.nl/web/static/lib/jquery/jquery.js:3231:49)
    at Object.<anonymous> (http://irodion.1systeem.nl/web/static/lib/jquery/jquery.js:3283:104)
```

Line 130 contains:

```
if (order && order.table.id != data.table_id) {
    order.transfer = true;
    order.destroy({'reason': 'abandon'});
}
```

Seems that order.table is empty here.

yelizariev:
fixed in 10.0 by @thomaspaulb

PR #519, opened 2017-12-05
[FIX] 10.0 - pos_multi_session_restaurant - dont error out if order.table is null

thomaspaulb:
Address #518 

No idea how order.table can be null though, @yelizariev any clue?

### DIFF
--- a/pos_multi_session_restaurant/static/src/js/pos_multi_session_restaurant.js
+++ b/pos_multi_session_restaurant/static/src/js/pos_multi_session_restaurant.js
@@ -127,7 +127,7 @@ odoo.define('pos_multi_session_restaurant', function(require){
                     return order.uid == data.uid;
                 });
             }
-            if (order && order.table.id != data.table_id) {
+            if (order && order.table && order.table.id != data.table_id) {
                 order.transfer = true;
                 order.destroy({'reason': 'abandon'});
             }


yelizariev:
Most obvious case: order was created before installing pos_restaurant module

blaggacao:
I can report a case, where I hit this, but pos_restaurant_module was installed all the time... It came up spontaniously...

blaggacao:
the data json is:
```json
message_id: 127
pos_id: 5
revsion_ID:1
uid: "00099-014-0289"
__proto__: Object
```

blaggacao:
-> `remove_order` ?

blaggacao:
... which would not be fixed by your commit.

blaggacao:
```
POS LONGPOLLING Biergarten2.2 pos.multi_session {"action": "remove_order", "data": {"pos_id":5,"revision_ID":1, "uid":"00088-014-0289", "message_ID":127}}
```

Seems this is the malicious call...

blaggacao:
I **guess** that it's because they where selling **while** updating the floor data, which is actually a rather valid case... At least the trace back should be handled and given instructions to cycle the session.

Imagine a big sales weekend and they need to quickly add more desks....

Can this be a reason? - After a complete cycle of all 14 tablets the error disappeared...

I'm not 100% confident on my analysis, still...

blaggacao:
In our case, the order is beeing **recently** created (table: null sic!)

```json
child {cid: "c4", attributes: {…}, _changing: false, _previousAttributes: {…}, changed: {…}, …}attributes: {client: null}changed: {}cid: "c4"collection: child {length: 17, models: Array(17), _byId: {…}, _events: {…}}creation_date: Sat Dec 09 2017 14:17:21 GMT-0500 (Eastern Standard Time) {constructor: ƒ}customer_count: 1finalized: falsefloor: undefinedinit_locked: falsems_info: {}ms_replace_empty_order: truename: "Pedido 00093-007-0230"new_order: falseorder_on_server: trueorderlines: child {length: 0, models: Array(0), _byId: {…}, _events: {…}}paymentlines: child {length: 0, models: Array(0), _byId: {…}, _events: {…}}pos: child {cid: "c2", attributes: {…}, _changing: false, _previousAttributes: {…}, changed: {…}, …}pos_session_id: 101revision_ID: 1saved_resume: undefinedscreen_data: {}selected_orderline: undefinedselected_paymentline: undefinedsequence_number: 231sequence_number_uniq: falsesession_id: 101table: nulltemporary: falseto_invoice: falseuid: "00093-007-0230"validation_date: "2017-12-09T18:43:14.004Z"_changing: false_events: {change: Array(1), change:sync: Array(1), all: Array(1)}_pending: false_previousAttributes: {client: null}__proto__: child
```

while the data has a timestamp from yesterday (and also a table):
```json
{pos_id: 5, uid: "00093-007-0230", ms_info: {…}, creation_date: "2017-12-08 22:52:01", customer_count: 1, …}amount_paid: 0amount_return: 0amount_tax: 0amount_total: 0creation_date: "2017-12-08 22:52:01"customer_count: 1fiscal_position_id: falsefloor: "Gr"floor_id: 3lines: []message_ID: 236ms_info: {created: {…}}name: "Pedido 00093-007-0230"new_order: falsepartner_id: falsepos_id: 5pos_session_id: 93revision_ID: 1sequence_number: 231sequence_number_uniq: falsestatement_ids: []table: "M15"table_id: 37uid: "00093-007-0230"user_id: 10__proto__: Object
```

blaggacao:
As a conclusion, I believe that the fix is not correct for this case and will produce stale orders which will stick around (forever?)

blaggacao:
What I had to do in order to not to risk anything:
```psql
DELETE FROM pos_multi_session_order WHERE order_uid='00093-007-0230';
```

I think this operation should be available from the UI...

yelizariev:
@blaggacao the solution could be a new module, that doesn't depend on pos_multi_session and can be used separately. See https://github.com/it-projects-llc/pos-addons/issues/528

yelizariev:
# Shall we revert this?
@GabbasovDinar 
Why the code removes order if tables are mismatch? What is the idea?
What is the possible workaround when table is unknown (see scenario above -- new table was added in one pos, but other poses are not reloaded)? Shall we
* remove order 
* or keep it
* or force user to reload POS?

The code:

```
+            if (order && order.table.id != data.table_id) {
+                order.transfer = true;
+                order.destroy({'reason': 'abandon'});
+            }
```
Corresponded commit: https://github.com/it-projects-llc/pos-addons/commit/7296132c6854dbbb4b4774566e2fa3b4ecec4b1c

blaggacao:
I guess as a **small** solution (without #528 ) the best would be to cycle (reload) the ~~session~~ POS.

PR #532, opened 2017-12-12
[IMP] refactoring of code, support for IOS 11.2

GabbasovDinar:


yelizariev (approved):

yelizariev:
- [x] Update README about this additional feature `` [IMP] show the guest number popup afte set empty table ``

By the way, you have to use tag ``[ADD]``, because you add feature rather than improve existing one

yelizariev (approved):
Code is reviewed

PR #557, opened 2018-01-15
11.0 pos debt notebook port

KolushovAlexandr:


pos_debt_notebook/__init__.py
```diff
@@ -1,16 +1,16 @@
-# -*- coding: utf-8 -*-
+
```
yelizariev: Fix empty lines

pos_debt_notebook/models.py
```diff
@@ -172,6 +172,14 @@ class PosConfig(models.Model):
              "without ordering new products. This is a workaround to the fact "
              "that Odoo needs to have at least one product on the order to "
              "validate the transaction.")
+    debt_type = fields.Selection([
+        ('debt', 'Display Debt'),
+        ('credit', 'Display Credit')
+    ], default='debt', string='Debt Type', help='Way to display debt value (label and sign of the amount). '
+                                                'In both cases debt will be red, credit - green')
```
yelizariev, line 172: This setting is used in partner form, so it has to be global and not per POS

yelizariev (changes requested):

pos_debt_notebook/models.py
KolushovAlexandr: @yelizariev in odoo 11.0 model pos.config.settings is included in pos.config, so i made the same changes. Where should they be then?

KolushovAlexandr: @yelizariev on the partner form it works correctly, but the partner form is reachable now only after installing account_invoicing, because the debendent module account is "Core mechanisms for the accounting modules. To display the menuitems, install the module account_invoicing."

yelizariev: Use ``res.config.settings`` for settings
Install module ``contacts`` to get partner views

Railia (approved):

ilmir-k (approved):

ilmir-k:
@yelizariev functional part has been reviewed, review the code please

yelizariev (approved):

pos_debt_notebook/__init__.py
```diff
@@ -1,16 +1,15 @@
...
-from openerp.tools.translate import _
+from odoo import SUPERUSER_ID
+from odoo import api
+from odoo.tools.translate import _
 from odoo.exceptions import UserError
 
 
 def pre_uninstall(cr, registry):
-    env = api.Environment(cr, SUPERUSER_ID, {})
+    env = api.Environment({})
```
yelizariev, line 12: WTF?

PR #544, opened 2017-12-25
10.0 pos debt notebook journal

KolushovAlexandr:

PR #570, opened 2018-01-31
[FIX] Debt history for Chinese localization

KolushovAlexandr:


ilmir-k:
@yelizariev we can merge it
unitest error could be fixed within  #544

ilmir-k (approved):

PR #575, opened 2018-02-09
[IMP] optimization and refactoring code

GabbasovDinar:


pos_multi_session/static/src/js/pos_multi_session.js
```diff
@@ -79,6 +89,9 @@ odoo.define('pos_multi_session', function(require){
             this.multi_session = false;
             this.ms_syncing_in_progress = false;
             this.ready.then(function () {
+                if (self.getUrlParameter('stringify-logs') === "1") {
+                    self.stringify_logs = true;
+                }
```
yelizariev: You don't need to wait for ready to compute this value

yelizariev (commented):
Seems fine, but you need to bump new version and add changelogs

ilmir-k (approved):

PR #586, opened 2018-02-15
[DEV] migration scripts

KolushovAlexandr:


pos_debt_notebook/migrations/10.0.5.0.0/post-migrate.py
```diff
@@ -0,0 +1,81 @@
...
+    # env holds all currently loaded models
+
+    debt_journals = env['account.journal'].search([('debt', '=', True)])
+
+    # a try to save data from ['product.template'].credit_product
+    if len(debt_journals) == 1:
+        # if there was only one debt journal, all debt products may be correlated with it
+        cr.execute('SELECT id, credit_product FROM product.template')
+        for record_id, old_value in cr.fetchall():
+            if old_value:
```
yelizariev: You have read credit_product before installation (i.e. move this to pre-XXX.py file).

Then only thing you can do there is make a copy of a column to read in in post-XXX.py file
All of it has to be done via pure SQL (not via normal orm)

pos_debt_notebook/migrations/10.0.5.0.0/post-migrate.py
```diff
@@ -0,0 +1,81 @@
...
+                new_value = (0, 0, debt_journals)
+                cr.execute('UPDATE product.template SET credit_product = new_value')
```
yelizariev: What? ``(0, 0, ...)`` is orm stuff, not for pure sql

yelizariev (changes requested):
PRE:

* Make temporal copy of  column credit_proudct

POST:

* update each product with non-empty values in temporarl column and set debt_journal there (first debt journal)

yelizariev (approved):

PR #601, opened 2018-02-22
pos_disable_payment: [CLEANUP] cleanup code and small refactorings

blaggacao:
**Note 1:** models.load_models can be replaced by one-liner models.load_fields to
	overload additional fields. Reduces LOCs and is more compliant with pos framework.

**Note 2:** (maybe related to Note 1)
From a note in the code:
```js
// At this point this.cashier has no rights settings added by module.
// Reset cashier to fix it
```
**Wrong** (from console):
```js
self.cashier
{groups_id: Array(29), pos_security_pin: false, allow_edit_price: true, allow_payments: true, allow_delete_order_line: true, …}
```

Reason of misunderstanding: get_cashier method tries to return directly from the local
client database, which in the case of rights that have changed will be written to db only
after the first set_cashier(). This is wired, but acceptable behaviour. It should be kept
in order to be fixed upstream.


blaggacao:
@yelizariev I'd love to see this merged, it would ease my rebasing a lot for what is to come in PRs.. :)

GabbasovDinar (approved):

ilmir-k (approved):
:+1:

PR #602, opened 2018-02-22
pos_disable_payment: [REF] respect framework recommendations

blaggacao:
I beg it's the same with a lot less code... :smile: 

- use start method where it's use is recomendend by the framework.
- untagle chrome-widget proxy, put on corresponding object and call directly,
	where applicable (OrderSelectorWidget)
- bind and render on corresponding objects instead of some parent proxy
	(NumpadWidget, ActionPadWidget, ProductScreenWidget).
- Reduce code duplications (ScreenWidget) by usage of `show` method on ProductScreenWidget

start - method recommendation (copied from code)
```js
    /**
     * Method called after rendering. Mostly used to bind actions, perform asynchronous
     * calls, etc...
     *
     * By convention, this method should return an object that can be passed to $.when()
     * to inform the caller when this widget has been initialized.
     *
     * @returns {jQuery.Deferred or any}
     */
    start: function() {
        return $.when();
    },
```

### DIFF
--- a/pos_disable_payment/static/src/js/pos_disable_payment.js
+++ b/pos_disable_payment/static/src/js/pos_disable_payment.js
@@ -52,11 +52,16 @@ odoo.define('pos_disable_payment', function(require){
         }
     });
 
-    chrome.Chrome.include({
-        init: function(){
-            this._super.apply(this, arguments);
+
+    chrome.OrderSelectorWidget.include({
+        start: function () {
             this.pos.bind('change:selectedOrder', this.check_allow_delete_order, this);
             this.pos.bind('change:cashier', this.check_allow_delete_order, this);
+            return this._super();
+        },
+        renderElement: function(){
+            this._super();
+            this.check_allow_delete_order();
         },
         check_allow_delete_order: function(){
             var user = this.pos.cashier || this.pos.user;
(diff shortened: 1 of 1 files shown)


yelizariev:
@blaggacao Thanks for working on it. Before we start checking it, I want to let you know that I have a plan to move this module to OCA https://github.com/OCA/pos/issues/232, because we don't sale this module anymore and it's quite common features, so we would like to stop maintaining it alone.
Do you think it's better to make your PRs there directly?

yelizariev:
I remember, that you have a bad feeling about OCA, but anyway

blaggacao:
@yelizariev I think it's better you delete the last comment, people might get hurt without a respectful contextualization... :wink: 

OK, here is the context for anyone who might feel offended https://docs.google.com/document/d/1WFsREZ0cygoiW-tTsZwrLCSkQ2DVa6aXMEvQncaUSKk/edit#heading=h.38l7uyb2sey4

Forked from: https://docs.google.com/document/d/1wwnu7oe5cDTyjZH3hdKkGLnF6fH4jPP3RLMPAhm3gjs/edit#

blaggacao:
OK, I would need to carry this one until the transition to OCA is made...

PR #603, opened 2018-02-22
pos_longpolling/pos_multi_session: [FIX] various longpolling issues

blaggacao:
This fix also dependes on a timeout in the odoo core bus module...


blaggacao:
Timeouts to protect deferred updates, if for any reason it takes longer than 3 seconds, the UI blocks with the turning circle suggesting you to take a coffee, which is pretty annoying.
Network is on: Just double check and update after the call so we have more timely feedback.
Randomness: Just avoid thundering heard in any possible given instance.

PR #607, opened 2018-02-22
pos_longpolling/pos_multi_session: [FIX] various longpolling issues

blaggacao:
This fix also dependes on a timeout in the odoo core bus module...

### DIFF
--- a/pos_multi_session/static/src/js/pos_multi_session.js
+++ b/pos_multi_session/static/src/js/pos_multi_session.js
@@ -744,7 +744,7 @@ odoo.define('pos_multi_session', function(require){
                     message: message,
                     dbname: session.db,
                     user_ID: self.pos.user.id
-                });
+                },{timeout:2500});
             };
             return send_it().fail(function (error, e) {
                 if (self.pos.debug){
@@ -794,6 +794,7 @@ odoo.define('pos_multi_session', function(require){
                     clearInterval(self.offline_sync_all_timer);
                     self.offline_sync_all_timer = false;
                 }
+                self.pos.longpolling_connection.network_is_on();
             });
         },
         destroy_removed_orders: function(server_orders_uid) {
@@ -839,7 +840,7 @@ odoo.define('pos_multi_session', function(require){
             var self = this;
             self.offline_sync_all_timer = setInterval(function(){
                 self.request_sync_all();
-            }, 5000);
+            }, 5000 + (Math.floor((Math.random()*10)+1)*1000));
         },
         no_connection_warning: function(){
             var warning_message = _t("No connection to the server. You can create new orders only. It is forbidden to modify existing orders.");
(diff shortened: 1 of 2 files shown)


blaggacao:
Timeouts to protect deferred updates, if for any reason it takes longer than 3 seconds, the UI blocks with the turning circle suggesting you to take a coffee, which is pretty annoying.
Network is on: Just double check and update after the call so we have more timely feedback.
Randomness: Just avoid thundering heard in any possible given instance.

blaggacao:
@yelizariev Could you have a look at this one? IIRC, it caused several hangups in our implementation while the server was not reachable...

blaggacao:
Hi guys, sorry, I don't want to pressure, but this is the last bit I would like to see included upstream before rebasing my own code... We surfaced this problem in the case of a unreachable network and it was quite annoying, you can imagine... I'd like to make sure this cannot happen or become an oversight in any rebase...

yelizariev (approved):
Looks good

blaggacao:
Wow, thank you! That was quick!! :smile:

pos_multi_session/static/src/js/pos_multi_session.js
```diff
@@ -794,6 +794,7 @@ odoo.define('pos_multi_session', function(require){
                     clearInterval(self.offline_sync_all_timer);
                     self.offline_sync_all_timer = false;
                 }
+                self.pos.longpolling_connection.network_is_on();
```
GabbasovDinar, line 797: @blaggacao Can you explain why this code is needed? Successful sending of data to the server does not guarantee to work of longpolling

GabbasovDinar:
@blaggacao We have the error

blaggacao:
Oh, sorry, must have been a regression. I have this code in production for some time now...

pos_multi_session/static/src/js/pos_multi_session.js
blaggacao: You're right. It was introduced in an attempt to make the UX more "reliable", because people have feared that when the sign is red, their order was not transferred and they typed the order again... Maybe the solution is in a improved UX concept... Example: A counter of pending local orders to synchronize. If synchronization is successful make it at least orange indicating that sending was possible but that there are no bidirectional guarantees, yet...

blaggacao (commented):
This is a good question... The answer might be a improvement of synchronization UX... I think we should reverse that line and once I rebase my own code, let's see if I can come up with something...

pos_multi_session/static/src/js/pos_multi_session.js
blaggacao: Or as this is a two arrow circle, it would be interesting to switch the colors of the two arrows independently... (upstream ok: lower error green, downstream ok: upper arrow green) I think that might actually be pretty helpful in a sometimes unstable WIFI environment...

blaggacao: Note: not sure how much you know about WIFI networking, but I've read a device can have full reception but close to no upstream connectivity under certain circumstances. Then you look at the signal and it looks 5 bars, but you hardly can push any data from your device...

yelizariev:
Fixed here https://github.com/it-projects-llc/pos-addons/pull/634/files
We are planning to make some refactoring there later

blaggacao:
Thanks! And again, sorry for the regression!

PR #616, opened 2018-02-25
10.0 pos debt notebook autopay recalculation, pos.order form

KolushovAlexandr:


pos_debt_notebook/models.py
```diff
@@ -101,7 +101,7 @@ def debt_history(self, limit=0):
                 )
                 for r2 in records:
                     r2['date'] = self._get_date_formats(r2['date'])
-                    r2['journal_code'] = self.env['account.journal'].search([('id', '=', r2['journal_id'][0])]).code
+                    r2['journal_code'] = self.env['account.journal'].browse(r2['journal_id'][0]).code
```
yelizariev, line 104: :+1:

pos_debt_notebook/static/src/js/pos.js
```diff
@@ -879,31 +887,33 @@ odoo.define('pos_debt_notebook.pos', function (require) {
...
-                    });
+                    $show_debt_history.on('click', this.show_debt_history_click.bind(self));
                 } else {
                     $show_debt_history.addClass('oe_hidden');
                     $show_debt_history.off();
                 }
             }
         },
-
+        show_debt_history_click: function(){
```
yelizariev: Make a separate PR for this update

yelizariev (changes requested):

pos_debt_notebook/models.py
Railia: @KolushovAlexandr An issue is produced when I first select Customer, and then when I cancel it, his values ​​remain in POs screen

yelizariev (approved):

PR #619, opened 2018-02-26
[FIX] incorrect debts representation after updating someones balance …

KolushovAlexandr:
…in a different pos


Railia:
@KolushovAlexandr  When you switch from one Customer to another in Show debt history, then the data of the other Client is displayed for a fraction of a second. But then everything works fine.

yelizariev (approved):
The code seems fine. I'm going to check features after merging

PR #620, opened 2018-02-26
[IMP] debt_history optimization

KolushovAlexandr:


Railia (approved):

pos_debt_notebook/models.py
```diff
@@ -89,25 +89,27 @@ def debt_history(self, limit=0):
...
-            data['debts'] = self.compute_debts_by_journals(r.id)
-            data['records_count'] = self.env['report.pos.debt'].search_count(domain)
-            data['partner_id'] = r.id
-            res.append(data)
-        return res
+        domain = [('partner_id', 'in', self.ids)]
+        records = self.env['report.pos.debt'].search_read(
+            domain=domain,
+            fields=fields,
+            limit=limit,
```
yelizariev: When limit is zero, search_read will return all records

yelizariev (changes requested):

yelizariev (changes requested):
It doesn't optimize anything. YOu still use ``.debt`` and ``compute_debts_by_journals`` per **each** record

PR #632, opened 2018-03-03
10.0-ref-pos_debt_notebook : clean up and refactoring

KolushovAlexandr:


pos_debt_notebook/models.py
```diff
@@ -173,12 +173,26 @@ def create_from_ui(self, partner):
 class PosConfig(models.Model):
     _inherit = 'pos.config'
 
+    @api.multi
+    def set_debt_type(self):
+        self.env["ir.config_parameter"].set_param("pos_debt_notebook.debt_type", self.debt_type)
```
yelizariev: This has to be in pos.config.settings. Otherwise it's confusing that debt_type is per POS, but it's noе

pos_debt_notebook/models.py
```diff
@@ -394,37 +408,6 @@ def _onchange_partner(self):
...
-        return {'debt_type': debt_type}
-
-    @api.multi
-    def set_debt_limit(self):
-        self.env["ir.config_parameter"].set_param("pos_debt_notebook.debt_limit", str(self.debt_limit))
-
-    @api.multi
-    def get_default_debt_limit(self, fields):
-        debt_limit = self.env["ir.config_parameter"].get_param("pos_debt_notebook.debt_limit", default=0)
-        return {'debt_limit': debt_limit}
```
yelizariev, line 425: removing debt_limit is ok, but the rest is not

yelizariev (changes requested):

yelizariev (commented):
Looks good

Railia (approved):

PR #633, opened 2018-03-03
[REM] remove Client from POS receipt

### LINKED ISSUE #426: [10.0-pos_debt_notebook]: Invading Receipt customization on pos.xml
In the pos.xml we have a nice, but harmful feature, which adds the customer to the ticket. But it breaks already customized tickets, and also uses a no translatable word. 

See:
https://github.com/it-projects-llc/pos-addons/blob/bd1ef7667b4bef2d330126acf1b1a74e86be6b85/pos_debt_notebook/static/src/xml/pos.xml#L150

yelizariev:
it breaks already customized tickets, and also uses a no translatable word.

closes #426

### DIFF
--- a/pos_debt_notebook/static/src/xml/pos.xml
+++ b/pos_debt_notebook/static/src/xml/pos.xml
@@ -143,14 +143,6 @@
     </t>
 
     <t t-extend="XmlReceipt">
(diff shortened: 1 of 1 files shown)


yelizariev:
test

yelizariev:
test

PR #638, opened 2018-03-06
[IMP] offline updating of all credits, not only via discounts ones

KolushovAlexandr:


yelizariev (changes requested):
Seems that it doesn't work with credit_product

yelizariev (dismissed):

pos_debt_notebook/data/product.xml
```diff
@@ -13,5 +13,16 @@
...
+            <field name="name">Credits top-up</field>
+            <field name="default_code">POS-DEBT</field>
+            <field name="sale_ok" eval="True"/>
+            <field name="available_in_pos" eval="True"/>
+            <field name="list_price">1</field>
+            <field name="type">service</field>
+            <field name="uom_id" ref="product.product_uom_unit"/>
+            <field name="description">This is a product for purchasing "Credits".</field>
+        </record>
+
```
yelizariev: It's not demo

pos_debt_notebook/models.py
```diff
@@ -266,6 +266,10 @@ def init_debt_journal(self):
         current_session.write({
             'statement_ids': statement,
         })
+        self.env.ref('pos_debt_notebook.product_credit_product').write({
+            'credit_product': debt_journal.id
+        })
+
```
yelizariev: It's not a demo

yelizariev (changes requested):

PR #647, opened 2018-03-09
[IMP] reload partners of opened orders

KolushovAlexandr:


yelizariev (changes requested):
It has to be completly remade.

* use Events instead of  calling method directly, so we can add more event handlers later if needed
* Reloading must be done not at the moment of load failing, but at the moment of restoring connection. Because debt might be loaded successfully, but we lost connection lately and partner balance was chnaged in another pos, while we lost notification about it through longpolling

pos_debt_notebook_sync/static/src/js/sync.js
```diff
@@ -14,10 +14,58 @@ odoo.define('pos_debt_sync', function(require){
...
         on_debt_updates: function(message){
             this.reload_debts(message.updated_partners);
-        }
+        },
+        _on_load_debts: function(debts){
+            var longpolling_connection = this.bus.longpolling_connection;
+            PosModelSuper.prototype._on_load_debts.apply(this, arguments);
+            if (longpolling_connection && !longpolling_connection.status){
+                longpolling_connection.send_ping();
+                this.trigger('restoredConnection');
```
yelizariev: This must be moved to pos_longpolling

pos_debt_notebook_sync/static/src/js/sync.js
```diff
@@ -14,10 +14,58 @@ odoo.define('pos_debt_sync', function(require){
...
+            }
+        },
+        _failed_load_debts: function(load_partner_ids){
+            var longpolling_connection = this.bus.longpolling_connection;
+            PosModelSuper.prototype._failed_load_debts.apply(this, arguments);
+            if (longpolling_connection && longpolling_connection.status){
+                longpolling_connection.send_ping();
+            }
```
yelizariev: Instead of redefining method, it's better to trigger event in main module and handle it here

pos_debt_notebook/static/src/js/pos.js
```diff
@@ -162,6 +161,11 @@ odoo.define('pos_debt_notebook.pos', function (require) {
             });
             this.trigger('updateDebtHistory', partner_ids);
         },
+        _failed_load_debts: function(load_partner_ids){
```
yelizariev: No need for this method. Update reload_debts to make postpone as amount of delay before loading, so you can replace
```
 setTimeout(_.bind(self.reload_debts, self,
                                load_partner_ids, 0, {"postpone": true, "shadow": false}), 3000);
```

with 
```
 self.reload_debts(load_partner_ids, 0, {"postpone": 4000, "shadow": false})
```

pos_debt_notebook_sync/static/src/js/sync.js
yelizariev: Or at least pos_longpolling has to trigger the event too (i.e. create new event , say change:http_connection)

pos_debt_notebook_sync/static/src/js/sync.js
```diff
@@ -14,10 +14,58 @@ odoo.define('pos_debt_sync', function(require){
...
+            var client_list_screen = this.gui.screen_instances.clientlist;
+            var partner = false;
+            if (client_list_screen && client_list_screen.clientlist_screen_is_opened()){
+                partner = client_list_screen.new_client || client_list_screen.pos.get_client();
+                partners_in_orders.push(partner.id);
+            }
+            if (partners_in_orders.length){
+                this.reload_debts(partners_in_orders, 0, {"postpone": true, "shadow": false}).then(function(){
+                    if (client_list_screen && partner){
+                        client_list_screen.update_debt_history([partner.id]);
```
yelizariev: update_debt_history is already called via updateDebtHistory event, isn't?

yelizariev (changes requested):

yelizariev (commented):
Nothing suspicious on the code

Railia:
This does not work. When I switch from offline to online, the data does not change. And the system shows errors.

pos_debt_notebook_sync/static/src/js/sync.js
```diff
@@ -14,10 +14,80 @@ odoo.define('pos_debt_sync', function(require){
...
+        _on_load_debts: function(debts){
+            var self = this;
+            var unsent_orders = _.filter(this.db.get_orders(), function(o){
+                return o.data.updates_debt;
+            });
+            if (unsent_orders){
+                var offline_order_partners = _.map(unsent_orders, function(o){
+                    return o.data.partner_id;
+                });
+                debts = _.filter(debts, function(deb){
```
yelizariev: Add comment explaining **why** you filter out it

pos_debt_notebook_sync/static/src/js/sync.js
```diff
@@ -14,10 +14,80 @@ odoo.define('pos_debt_sync', function(require){
...
+                }
+            });
+            var client_list_screen = this.gui.screen_instances.clientlist;
+            var partner = false;
+            if (client_list_screen && client_list_screen.clientlist_screen_is_opened()){
+                partner = client_list_screen.new_client || client_list_screen.pos.get_client();
+                partners_in_orders.push(partner.id);
+            }
+
+            var pushed_orders = $.Deferred();
```
yelizariev: No need for this. push_order already returns object that can be used to call ``then``

pos_debt_notebook_sync/static/src/js/sync.js
```diff
@@ -14,10 +14,80 @@ odoo.define('pos_debt_sync', function(require){
...
+            _.each(this.db.get_orders(), function(o){
+                if (o.data.updates_debt && o.data.partner_id){
+                    partners_in_orders.push(o.data.partner_id);
+                }
+            });
+            var client_list_screen = this.gui.screen_instances.clientlist;
+            var partner = false;
+            if (client_list_screen && client_list_screen.clientlist_screen_is_opened()){
+                partner = client_list_screen.new_client || client_list_screen.pos.get_client();
+                partners_in_orders.push(partner.id);
```
yelizariev: Rename ``partners_in_orders``, because you put there partner from clien list.
It could be ``partners_to_reload`` or something

yelizariev (dismissed):

yelizariev (commented):
It should work

yelizariev (approved):

pos_debt_notebook/doc/changelog.rst
```diff
@@ -8,6 +8,8 @@
 - **FIX:** Debt history for Chinese localization
 - **FIX:** Error when user doesn't have timezone
 - **FIX:** user_id was always replaced to Administrator in pos.order
+- **IMP:** Updating opened sessions partners debts/credits after restoring the internet connection
+- **ADD:** Autopay button in POS
```
yelizariev: There is no Autopay button in POS (i.e. in /pos/web). There is no such button at all.

pos_debt_notebook/doc/changelog.rst
```diff
@@ -8,6 +8,8 @@
 - **FIX:** Debt history for Chinese localization
 - **FIX:** Error when user doesn't have timezone
 - **FIX:** user_id was always replaced to Administrator in pos.order
+- **IMP:** Updating opened sessions partners debts/credits after restoring the internet connection
```
yelizariev: ``opened sessions partners debts/credits `` -- what is it? Did you mean opened orders?
Anyway, you must remake that sentence, because you have here ``adjective + noun + noun + noun``

yelizariev (changes requested):

Railia (approved):

PR #650, opened 2018-03-13
[FIX] change for credits via discounts, that cause standard payments …

KolushovAlexandr:
…instead of manual credit updates

### DIFF
--- a/pos_debt_notebook/static/src/js/pos.js
+++ b/pos_debt_notebook/static/src/js/pos.js
@@ -427,11 +427,15 @@ odoo.define('pos_debt_notebook.pos', function (require) {
                     var discount_credits = Math.max(0, order.get_summary_for_discount_credits());
                     var percentage = 0;
                     var orderlines = order.get_orderlines();
+                    var old_price = 0;
                     _.each(orderlines, function(ol){
                         if (discount_credits > 0.00001) {
-                            percentage = Math.min(Math.max(( discount_credits / ol.get_unit_price() ) * 100, 0), 100);
+                            old_price = ol.get_price_with_tax();
+                            percentage = Math.min(Math.max(( discount_credits / old_price ) * 100, 0), 100);
                             ol.set_discount(percentage);
-                            discount_credits -= ol.get_unit_price() * (percentage) / 100;
+                            discount_credits -= old_price - ol.get_price_with_tax();
+                        } else {
+                            return;
                         }
                     });
                 }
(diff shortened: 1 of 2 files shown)


pos_debt_notebook/static/src/js/pos.js
```diff
@@ -424,14 +424,19 @@ odoo.define('pos_debt_notebook.pos', function (require) {
...
-                    }, 0);
+                    var sum = Math.max(0, order.get_summary_for_discount_credits());
                     var percentage = ( sum / order_total ) * 100;
                     var orderlines = order.get_orderlines();
                     _.each(orderlines, function(ol){
                         ol.set_discount(percentage);
                     });
+                    // fixing of the last product discount to prevent price rounding errors after discount setting
+                    // it causes adding a payment for discount credits not via Manual Credit Updates but via standard methods
+                    var diff_of_rounding = order.get_total_paid() - order.get_total_with_tax() - sum;
```
yelizariev: Try to recompute percentage in the lines above instead (i.e. recompute percentage for each line). The way you did it now looks ugly and difficult to understand

``ol.set_discount(( 1 - ( ol.get_price_with_tax() + diff_of_rounding )/ol.get_unit_price()) * 100);``

yelizariev (changes requested):

pos_debt_notebook/static/src/js/pos.js
```diff
@@ -424,13 +424,18 @@ odoo.define('pos_debt_notebook.pos', function (require) {
...
+                    var remained_amount = order_total;
+                    var remained_sum = sum;
                     var orderlines = order.get_orderlines();
+                    var diff_of_rounding = 0;
                     _.each(orderlines, function(ol){
+                        diff_of_rounding = order.get_total_paid() - order.get_total_with_tax() - sum + remained_sum;
+                        percentage = ( ( remained_sum - diff_of_rounding ) / remained_amount ) * 100;
                         ol.set_discount(percentage);
+                        remained_sum -= ol.get_unit_price() * (percentage) / 100;
+                        remained_amount -= ol.get_unit_price();
```
yelizariev: Can you call ``get_total_with_tax`` after each discount instead of introducing new variables and making computtation that difficult to understand

yelizariev (changes requested):

yelizariev (approved):
The code updates look fine

Railia (approved):

PR #652, opened 2018-03-13
[IMP] not block UI after first load data from server

GabbasovDinar:


pos_multi_session/static/src/js/pos_multi_session.js
```diff
@@ -700,7 +701,10 @@ odoo.define('pos_multi_session', function(require){
                 return d.promise();
             }
 
-            framework.blockUI();
+            if (this.first_load_data_from_server) {
+                framework.blockUI();
```
yelizariev: Make it as a part of initial black screen (the same as you did with @KolushovAlexandr  for pos_debt_notebook)

yelizariev (changes requested):

pos_multi_session/static/src/js/pos_multi_session.js
GabbasovDinar: @yelizariev 
no-god-please-onyf30

pos_multi_session/static/src/js/pos_multi_session.js
```diff
@@ -124,10 +124,39 @@ odoo.define('pos_multi_session', function(require){
...
+            var loaded = new $.Deferred();
+            var done = new $.Deferred();
+
+            if (res) {
+                res.then(function() {
+                    loaded.resolve();
+                });
+            } else {
+                loaded.resolve();
+            }
```
yelizariev: You can use $.when(res). No need for if-then. Neither for ``loaded``

pos_multi_session/static/src/js/pos_multi_session.js
```diff
@@ -124,10 +124,39 @@ odoo.define('pos_multi_session', function(require){
...
+                self.multi_session.bind_ms_connection_events();
+            });
+        },
+        after_load_server_data: function() {
+            var self = this;
+            if (!this.config.multi_session_id){
+                return PosModelSuper.prototype.after_load_server_data.apply(this, arguments);
+            }
+
+            this.multi_session = new exports.MultiSession(self);
```
yelizariev: It's better to do it in ``initilalize`` or ``init`` method

yelizariev (changes requested):

pos_multi_session/static/src/js/pos_multi_session.js
```diff
@@ -690,8 +723,13 @@ odoo.define('pos_multi_session', function(require){
             });
         },
         sync_all: function(data) {
+            if (this.first_load_data_from_server) {
```
yelizariev: Add new variable to ``request_sync_all``, say ``options`` and pass it to here when data are received. Try to make code without "side effects".

pos_multi_session/static/src/js/pos_multi_session.js
```diff
@@ -710,8 +748,26 @@ odoo.define('pos_multi_session', function(require){
...
+            });
+
+            return done;
+        },
+        update_orders: function(data) {
+            var server_orders_uid = [];
+            var self = this;
+
+            data.orders.forEach(function (item, index) {
+                self.pos.ms_on_update(item, true);
```
yelizariev: It's a code duplicate. Try to use variable in ``sync_all`` method to define do we need to use timeout or not

yelizariev (changes requested):
Not making timeout for rendering on initial sync is a good idea. Are you sure, that those order are not rendered on "Black screen stage"?

pos_multi_session/static/src/js/pos_multi_session.js
GabbasovDinar: @yelizariev it's not possible. We cannot check this.config.multi_session_id condition since this.config available after completing the load_server_data function only. The deferred object this.ready get 'resolve' object after completing the after_load_server_data function so I think it's right to initialize the multi-session here.

pos_multi_session/static/src/js/pos_multi_session.js
GabbasovDinar: @yelizariev why? in this case, we need send the parameter to the server and get a result with this parameter for sending to sync_all function

yelizariev: No need to send anything to the server to call method with extra argument

pos_multi_session/static/src/js/pos_multi_session.js
```diff
@@ -740,13 +724,21 @@ odoo.define('pos_multi_session', function(require){
 
             this.q = $.when();
             var done = $.Deferred();
-            data.orders.forEach(function (item, index) {
-                self.q = self.q.then(function(){
+
+            if (this.first_load_data_from_server) {
```
yelizariev: Still bad idea. Use extra argument instead

pos_multi_session/static/src/js/pos_multi_session.js
```diff
@@ -124,10 +124,27 @@ odoo.define('pos_multi_session', function(require){
...
+                self.multi_session.bind_ms_connection_events();
+            });
+        },
+        after_load_server_data: function() {
+            var self = this;
+            if (!this.config.multi_session_id){
+                return PosModelSuper.prototype.after_load_server_data.apply(this, arguments);
+            }
+            this.multi_session = new exports.MultiSession(self);
+            var res = PosModelSuper.prototype.after_load_server_data.apply(self, arguments);
```
yelizariev: You can remake this code without calling ``after_load_server_data`` in two places (hint: move ``var res = .... `` at the beginning.

pos_multi_session/static/src/js/pos_multi_session.js
```diff
@@ -676,20 +696,26 @@ odoo.define('pos_multi_session', function(require){
...
+            options = options || {}
             this.on_syncing = true;
             var data = {run_ID: this.pos.multi_session_run_ID};
-            var message = {'action': 'sync_all', data: data, 'uid': order_uid};
+            var message = {'action': 'sync_all', data: data};
+            if (options.uid) {
+                message.uid = options.uid;
+            }
+            if (options.first_load) {
+                message.first_load = options.first_load;
```
yelizariev: No need to send this argument to the server. Put extra arg to method ``send`` instead.

yelizariev (changes requested):

GabbasovDinar:
@yelizariev DONE

pos_multi_session/static/src/js/pos_multi_session.js
```diff
@@ -742,7 +773,11 @@ odoo.define('pos_multi_session', function(require){
             });
         },
         _debug_send_number: 0,
-        send: function(message, address){
+        send: function(message, options){
+            options = options || {};
+            if (options.uid) {
+                message.uid = options.uid;
+            }
```
yelizariev: uid is specific for sync_all only. Move it to request_sync_all

yelizariev (approved):
Almost ready.

pos_multi_session/static/src/js/pos_multi_session.js
```diff
@@ -700,26 +719,38 @@ odoo.define('pos_multi_session', function(require){
...
-                    return delay(100);
+                    server_orders_uid.push(item.data.uid);
                 });
-                server_orders_uid.push(item.data.uid);
-            });
-
-            self.q.then(function() {
-                framework.unblockUI();
-            });
+            } else {
```
yelizariev, line 771: Option name must describe how method shall be processed and not in which cases (when) it's used.

yelizariev (changes requested):

pos_multi_session/static/src/js/pos_multi_session.js
yelizariev: rename ``first_load`` to immediate_rerendeing,  or sync_update or similar

pos_multi_session/static/src/js/pos_multi_session.js
```diff
@@ -140,7 +140,7 @@ odoo.define('pos_multi_session', function(require){
                 var progress = (self.models.length - 0.5) / self.models.length;
                 self.chrome.loading_message(_t('Sync Orders'), progress);
 
-                return self.multi_session.request_sync_all({'first_load': true}).then(function() {
+                return self.multi_session.request_sync_all({'immediate_rerendeing': true}).then(function() {
```
yelizariev: fix spelling ``immediate_rerendeing``

yelizariev: Also, the name is not ideal as you said, that on initial black screen orders are not rendered at all

yelizariev (changes requested):
Alternative name: update_orders_immediatly

pos_multi_session/static/src/js/pos_multi_session.js
GabbasovDinar: @yelizariev What name can I use for this variable?

GabbasovDinar: @yelizariev update_orders_immediatly or immediate_rerendeing?

yelizariev: You still want to use wrong spelling **rerendeing**? :smile:

yelizariev: Choose yourself

yelizariev (approved):
Code is fine now

yelizariev (approved):
I looked through. It looks fine

ilmir-k (approved):

PR #653, opened 2018-03-15
[NEW] pos_barcode_sync

KolushovAlexandr:


pos_barcode_sync/static/src/js/pos_barcode_sync.js
```diff
@@ -0,0 +1,56 @@
...
+    var longpolling = require('pos_longpolling');
+    var gui = require('point_of_sale.gui');
+
+    var _t = core._t;
+
+    var PosModelSuper = models.PosModel;
+    models.PosModel = models.PosModel.extend({
+        initialize: function(){
+            PosModelSuper.prototype.initialize.apply(this, arguments);
+            this.bus.add_channel_callback("pos_barcode_sync", this.update_partners, this);
```
yelizariev, line 17: I cannot where is update_partners defined

pos_barcode_sync/static/src/js/pos_barcode_sync.js
```diff
@@ -0,0 +1,56 @@
...
+            this.on('updatePartnerBarcode', function(partner_id){
+                session.rpc('/barcode/update', {message: 'Update Partner Barcode', partner_id: partner_id}, {shadow : true});
```
yelizariev, line 19: Barcode can be changed not via POS. Use base.action rule instead.
wrong

yelizariev (changes requested):

yelizariev:
Send it to addons-dev

PR #654, opened 2018-03-15
[FIX] error on empty multi-session field

KolushovAlexandr:


Railia:
@KolushovAlexandr I chose table 2 and opened two sessions, adding the product both sessions. And there is no synchronization. On the table 3 everything is synchronising.

Railia (changes requested):
@KolushovAlexandr The issue with user settings. Disable Allow discount, Edit price, but still it is available for the user

Railia (approved):

Railia:
@yelizariev Could you help us to merge this pull request since these lints won't be fixed

PR #658, opened 2018-03-25
[REF] refactoring code and speed optimization

GabbasovDinar:


yelizariev (approved):
orderline_change is redefined as planned

ilmir-k (approved):

PR #663, opened 2018-03-30
disabe autopay for some cases

KolushovAlexandr:


pos_debt_notebook/models.py
```diff
@@ -261,7 +261,7 @@ def init_debt_journal(self):
                                                 'debt_dummy_product_id': self.env.ref('pos_debt_notebook.product_pay_debt').id,
                                                 'debt_limit': default_debt_limit,
                                                 'pos_cash_out': True,
-                                                'credits_autopay': True,
+                                                'credits_autopay': False,
```
yelizariev: I think we should keep it True, i.e. activate feature in new installation

yelizariev (changes requested):
Add copyrights and license to all files, please

pos_debt_notebook/static/src/js/pos.js
```diff
@@ -727,7 +734,7 @@ odoo.define('pos_debt_notebook.pos', function (require) {
                 return;
             }
             this._super();
-            if (!order.get_paymentlines().length){
+            if (!order.get_paymentlines().length || order.has_credit_product()){
```
yelizariev: ``render_paymentlines`` probably need to be updated. It looks strange, that it makes update only when there are no lines

yelizariev (approved):
Code is ok

ilmir-k (approved):

PR #668, opened 2018-04-03
[FIX] TypeError: self.new_client is null when `Pay Full Debt

KolushovAlexandr:


ilmir-k (approved):

ilmir-k:
@yelizariev it can be merged

yelizariev (approved):

PR #672, opened 2018-04-07
[10.0][FIX] pos_multi_session: sync add_orderline

ata2001:
This change was needed to make [my addon](https://www.odoo.com/apps/modules/10.0/pos_merge_orders/) work with the pos_multi_session module.

### DIFF
--- a/pos_multi_session/static/src/js/pos_multi_session.js
+++ b/pos_multi_session/static/src/js/pos_multi_session.js
@@ -4,6 +4,7 @@
  * Copyright 2016-2018 Dinar Gabbasov <https://it-projects.info/team/GabbasovDinar>
  * Copyright 2017 Kolushov Alexandr <https://it-projects.info/team/KolushovAlexandr>
  * Copyright 2017 David Arnold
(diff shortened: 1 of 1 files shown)


pos_multi_session/static/src/js/pos_multi_session.js
```diff
@@ -493,6 +494,10 @@ odoo.define('pos_multi_session', function(require){
                 self.ms_update();
             });
         },
+        add_orderline: function(line){
+            OrderSuper.prototype.add_orderline.apply(this, arguments);
```
GabbasovDinar, line 498: @ata2001 Why do you think that this function is necessary? Сan you describe in more detail? 
Thanks

GabbasovDinar: @ata2001 Sorry, I saw your description. I'll check it out soon.

yelizariev:
You can trigger the event in your module. 
If the event is not needed without your module, it means that we decrease performance for nothing

yelizariev:
Close as no feedback from @ata2001

PR #673, opened 2018-04-09
[FIX] don't remove other orders if sync the current order

GabbasovDinar:


Railia:
@GabbasovDinar Orders from other tables do not disappear, but the synchronization error appears

Railia (approved):

GabbasovDinar:
@Railia  this PR does not resolve the synchronization error. The PR solves the error of deleting orders only

ilmir-k (approved):

PR #674, opened 2018-04-10
[FIX] timeout of update

GabbasovDinar:
The timeout of 30 seconds allows fixing the revision error when the connection is slow


Railia (changes requested):
@GabbasovDinar the same error is reproducing: synchronization error

ilmir-k (approved):

Railia (approved):

PR #676, opened 2018-04-15
 [FIX] order in undefined pos_multi_session_restaurant

GabbasovDinar:


pos_print_method/__manifest__.py
```diff
@@ -12,7 +12,7 @@
     "support": "apps@it-projects.info",
     "website": "https://it-projects.info/team/GabbasovDinar",
     "license": "LGPL-3",
-    "price": 69.00,
+    "price": 34.00,
```
yelizariev, line 15: WTF

yelizariev:
fixed in #675

PR #682, opened 2018-04-26
[FIX] Printing orders after connection recovery of Network Printer.

GabbasovDinar:
The fixes for hw_printer_network module

### DIFF
--- a/hw_printer_network/README.rst
+++ b/hw_printer_network/README.rst
@@ -1,3 +1,7 @@
+.. image:: https://img.shields.io/badge/license-LGPL--3-blue.png
+   :target: https://www.gnu.org/licenses/lgpl
+   :alt: License: LGPL-3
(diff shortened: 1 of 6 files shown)


ilmir-k (approved):

yelizariev (approved):
ok
But we need expect following scenarios:

* printer was offline long time (e.g. at night), and then it starts print all lost tickets

Admin-Ever:
@GabbasovDinar  Dear I was trying to find a solution with the network printers whenever one printer is down of the order printers , the you cant print anymore it looks like there s kind the receipt are stuck somewhere waiting for the printer 
Expected behavior is to keep on trying to print that printer but continue with the other receipt
any quick tip i m ready to help with anything needed just i need a hint or a solution if exist 
thank you

PR #683, opened 2018-04-26
[FIX] sync error during fast change the order

GabbasovDinar:


yelizariev (approved):

yelizariev:
@ilmir-k use this commit message on merging

```
[FIX] error on removing right after updating an order

request for order removing must be done after or instead of the request for updating.

Without this update, the script could send a request for an update after sending a request for removing the order. The scenario: 

* update order -- timer start
* remove order -- the request is sent immediately
* timer ends -- request for updating is sent
```

ilmir-k (approved):

ilmir-k:
@GabbasovDinar fix the remaining lints

GabbasovDinar:
@ilmir-k DONE

ilmir-k:
@yelizariev I can't merge it due to travis warnings

yelizariev:
Those errors have to be fixed

yelizariev:
While the error may be simple and could be ignored, it actually blocks all other checks which might find essential errors

GabbasovDinar:
@yelizariev @ilmir-k DONE

PR #677, opened 2018-04-19
[FIX] order is null

GabbasovDinar:

PR #688, opened 2018-04-30
10.0 fix pos invoice pay

yelizariev:


yelizariev:
Closed in favor of #677

PR #691, opened 2018-05-09
Fixes #690

### LINKED ISSUE #690: [10.0] pos_debt_notebook: The generated invoices use the payable account instead of the receivable account.
The invoices created thought the wizard end up using the wrong account and payment term, they are customer invoices, but they are created with the payable account of the partner.

This is because the wizard calls the create method without the 'type' value, which defaults to 'out_invoice', but the defaults are run at the end of the creation, and the onchange method for the partner_id, which in turn sets the account and the payment term is run first, so the type at that moment is False, and the onchange sets the account for a vendor invoice.

I'll be making a PR to fix this.

carlosrve:
Fix issue #690

### DIFF
--- a/pos_debt_notebook/wizard/pos_credit_invoices.py
+++ b/pos_debt_notebook/wizard/pos_credit_invoices.py
@@ -66,6 +66,7 @@ class PosCreditInvoices(models.TransientModel):
             if not line.amount:
                 continue
             self.env['account.invoice'].create({
+                'type': 'out_invoice',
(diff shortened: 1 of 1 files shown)


Ramil-Mukhametzyanov (approved):
[Approved] Invoices created in the "Generate POS Credit Invoices" wizard affect the company account after payment

yelizariev:
We will merge in a new PR, because we need to update changelog

yelizariev:
Finally merged in #777 
Thank you!

PR #700, opened 2018-06-06
[IMP] pos_multi_session: new option for POS in offline mode

GabbasovDinar:
New option to save orderlines added to existing orders in offline mode


yelizariev (changes requested):
Blocked until tests are done

Railia (approved):

pos_multi_session/models/pos_multi_session_models.py
```diff
@@ -24,6 +24,7 @@ class PosConfig(models.Model):
     multi_session_accept_incoming_orders = fields.Boolean('Accept incoming orders', default=True)
     multi_session_replace_empty_order = fields.Boolean('Replace empty order', default=True, help='Empty order is deleted whenever new order is come from another POS')
     multi_session_deactivate_empty_order = fields.Boolean('Deactivate empty order', default=False, help='POS is switched to new foreign Order, if current order is empty')
+    multi_session_block_exist_order = fields.Boolean('Block existing order', default=True, help='Block the existing order if the POS does not have a connection to the server')
```
yelizariev, line 27: Order is not blocked. But updates of existing order are ignored

yelizariev (commented):
It shall not be an option. Always try to send updates to server (i.e. multi_session_block_exist_order is always false)

GabbasovDinar:
was fixed here https://github.com/it-projects-llc/pos-addons/pull/747

PR #708, opened 2018-06-27
[FIX] compatibility issue between pos_disable_payments & pos_keyboard

ssaid:
Add new funtion to control whether it is allowed the operation

### DIFF
--- a/pos_keyboard/static/src/js/pos.js
+++ b/pos_keyboard/static/src/js/pos.js
@@ -81,14 +81,14 @@ function pos_keyboard_widgets(instance, module){
 
         is_allowed: function(ttype){
             // pos_disable_payment installed?
-            if (!window.posmodel.config.hasOwnProperty('allow_discount')) {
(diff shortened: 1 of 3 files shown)


pos_keyboard/static/src/js/pos.js
```diff
@@ -78,6 +78,20 @@ function pos_keyboard_widgets(instance, module){
         reset_action_callback: function(){
             this.action_callback = undefined;
         },
+
+        is_allowed: function(ttype){
+            // pos_disable_payment installed?
+            if (!window.posmodel.config.hasOwnProperty('allow_discount')) {
```
GabbasovDinar: use ``this.pos`` instead of ``window.posmodel``

Ramil-Mukhametzyanov (approved):
If 'Allow discount' is deactivated, the user can not set a discount for a customer.
If 'Allow edit price' is deactivated, the user can not change the prices of the products for a customer.

PR #716, opened 2018-07-18
10.0 pos ms old orders in new session

KolushovAlexandr:


KolushovAlexandr:
changes were merged into another branch/pr

PR #718, opened 2018-07-18
10.0 pos ms renamings

KolushovAlexandr:


yelizariev:
@KolushovAlexandr did you run *external tests* after renaming? Could  you make a PR with renamings only?

KolushovAlexandr:
@yelizariev external tests passed, PR with renamings only https://github.com/it-projects-llc/pos-addons/pull/720

yelizariev:
merged in #720

PR #727, opened 2018-07-30
:zap: closing a warning message after connection restoring

KolushovAlexandr:


yelizariev (changes requested):
It will not work when translation is activated

yelizariev (changes requested):
Update changelog please

PR #731, opened 2018-08-07
10.0 ms sync conflicts SYNC ALL REF

KolushovAlexandr:


yelizariev:
- [x] variable ``sync_all`` is specified and method args, but not used

yelizariev:
- [x] ``sync_all`` method has destroy_removed_orders. Can it be moved to ``updates_from_server``?

yelizariev:
- [x] update changelog

pos_multi_session/static/src/js/pos_multi_session.js
```diff
@@ -276,9 +277,17 @@ odoo.define('pos_multi_session', function(require){
...
+                if (action === 'sync_all') {
                     this.message_ID = data.message_ID;
-                    this.ms_update_order(order, data);
+                    _.each(data.orders, function(received_data){
+                        order = self.get('orders').find(function(item){
+                            return item.uid === received_data.data.uid;
+                        }) || false;
+                        self.ms_update_order(order, received_data.data);
+                    });
+                    this.multi_session.destroy_removed_orders(_.map(data.orders, function(o){
```
yelizariev: Do you remove all orders here?

yelizariev (changes requested):

pos_multi_session/static/src/js/pos_multi_session.js
yelizariev: ok, you don't

pos_multi_session/static/src/js/pos_multi_session.js
```diff
@@ -805,54 +843,23 @@ odoo.define('pos_multi_session', function(require){
...
-
-            if (!options.sync_current_order) {
-                this.destroy_removed_orders(server_orders_uid);
-            }
-            self.q.then(function() {
-                done.resolve();
-            });
-
-            return done;
+            this.destroy_removed_orders(server_orders_uid);
```
yelizariev: It's already called in updates_from_server

yelizariev (changes requested):

pos_multi_session/static/src/js/pos_multi_session.js
```diff
@@ -805,54 +843,23 @@ odoo.define('pos_multi_session', function(require){
...
-                });
-            }
+            var data = message.data;
+            data.orders.forEach(function (item) {
+                server_orders_uid.push(item.data.uid);
+            });
 
             this.pos.pos_session.order_ID = data.order_ID;
+            var sequence_number = this.pos.pos_session.sequence_number;
+            this.pos.pos_session.sequence_number = Math.max(Boolean(sequence_number) && sequence_number, data.order_ID + 1 || 1);
```
yelizariev: Why there is no the same code in updates_from_server?

yelizariev (changes requested):
Could you make more clean work work with .offline_orderline? For example mark all new lines as offline and unmark once it’s send successfully to server

pos_multi_session/static/src/js/pos_multi_session.js
```diff
@@ -347,19 +347,11 @@ odoo.define('pos_multi_session', function(require){
...
-                    lines: false,
-                    multiprint_resume: data.multiprint_resume,
-                    new_order: false,
+                var json = _.extend(data, {
                     order_on_server: true,
-                    table_id: data.table_id,
-                    floors_id: data.floor_id
-                };
-                order = this.ms_create_order({ms_info:data.ms_info, revision_ID:data.revision_ID, data:data, json:json});
+                    multiprint_resume: data.multiprint_resume,
```
yelizariev: this lins is useless

pos_multi_session/static/src/js/pos_multi_session.js
```diff
@@ -172,6 +173,7 @@ odoo.define('pos_multi_session', function(require){
 
                 var load_sync_all_request = function(){
                     return self.multi_session.request_sync_all({'immediate_rerendering': true}).then(function() {
+                        self.is_loaded = true;
```
yelizariev, line 176: use ``pos.ready....`` instead

pos_multi_session/static/src/js/pos_multi_session.js
```diff
@@ -392,19 +394,32 @@ odoo.define('pos_multi_session', function(require){
                 } else if (dline.is_changed) {
                     line.apply_ms_data(dline);
                 }
+                line.offline_orderline = false;
```
yelizariev, line 396: Consider to update this variable via options (like not_render)

pos_multi_session/static/src/js/pos_multi_session.js
```diff
@@ -338,22 +347,15 @@ odoo.define('pos_multi_session', function(require){
...
-                    new_order: false,
+                var json = _.extend(data, {
                     order_on_server: true,
-                    table_id: data.table_id,
-                    floors_id: data.floor_id
-                };
-                order = this.ms_create_order({ms_info:data.ms_info, revision_ID:data.revision_ID, data:data, json:json});
+                });
+                order = this.ms_create_order({ms_info:data.ms_info, revision_ID:data.revision_ID, json: json});
+                order.init_locked = false;
```
yelizariev: You don't need it here if you remove it from ms_create_order

yelizariev (changes requested):

pos_multi_session/static/src/js/pos_multi_session.js
KolushovAlexandr: @yelizariev they are partially correlated, but I see no way to make offline_orderline depends on not_render

pos_multi_session/static/src/js/pos_multi_session.js
KolushovAlexandr: @yelizariev 
// init_locked blocks execution of save_to_db
// the order is unlocked at the end of ms_update_order
probably the removing might causes some problems?

ilmir-k:
@yelizariev the changes requested are fixed I suppose

pos_multi_session/static/src/js/pos_multi_session.js
yelizariev: I meant to set ``offline_orderline`` in the same way as ``not_render`` (via options to don't set it directly)

pos_multi_session/static/src/js/pos_multi_session.js
yelizariev: no

yelizariev:
@ilmir-k not really

GabbasovDinar (approved):

yelizariev (approved):

PR #737, opened 2018-08-18
9.0 debt backport

yelizariev:


yelizariev:
We will upload version from 10 branch and fix it instead of trying to apply commits from there

PR #739, opened 2018-08-24
11.0 pos product available negative port

KolushovAlexandr:


yelizariev (approved):

ilmir-k (approved):
Approved, but the tests have not been passed for some reason

ilmir-k:
@yelizariev can we merge it?

yelizariev:
:ambulance: has to be used when *before the commit it didn't work correctly* which is not the case. Please fix

PR #744, opened 2018-08-28
:ambulance: fixed wrong calculation of partner debts

KolushovAlexandr:


yelizariev (changes requested):
Could you fix the previous implementation, because it's quicker way to compute (it makes only one sql request to db instead of num(partners)

pos_debt_notebook/models.py
```diff
@@ -31,19 +31,12 @@ def _compute_debt(self):
             for partner_id, val in self._cr.fetchall():
                 res[partner_id] += val
 
-        statements = self.env['account.bank.statement'].search(
-            [('journal_id', 'in', [j.id for j in debt_journal]), ('state', '=', 'open')])
+        orders = self.env['pos.order'].search([('partner_id', 'in', self.ids)])
+        for o in orders:
+            for s in o.statement_ids:
+                if s.journal_id.debt:
+                    res[o.partner_id.id] += s.amount
```
yelizariev: Why don't you check statement state?

pos_debt_notebook/models.py
```diff
@@ -31,19 +31,28 @@ def _compute_debt(self):
             for partner_id, val in self._cr.fetchall():
                 res[partner_id] += val
 
-        statements = self.env['account.bank.statement'].search(
-            [('journal_id', 'in', [j.id for j in debt_journal]), ('state', '=', 'open')])
-
+        orders = self.env['pos.order'].search([('partner_id', 'in', self.ids)])
+        statements_ids = [st.id for o in orders for st in o.statement_ids]
```
yelizariev: Can you use ``mapped`` intead?

pos_debt_notebook/models.py
```diff
@@ -31,19 +31,28 @@ def _compute_debt(self):
...
         if statements:
             self._cr.execute(
-                """SELECT l.partner_id, SUM(l.amount)
-                FROM account_bank_statement_line l
-                WHERE l.statement_id IN %s AND l.partner_id IN %s
-                GROUP BY l.partner_id
+                """SELECT rp.id, SUM(absl.amount)
+                FROM pos_order AS po,
+                     res_partner AS rp,
+                     account_bank_statement_line AS absl
```
yelizariev: three FROM without JOIN looks like a bad idea

pos_debt_notebook/models.py
```diff
@@ -31,19 +31,28 @@ def _compute_debt(self):
...
                 res[partner_id] += val
 
-        statements = self.env['account.bank.statement'].search(
-            [('journal_id', 'in', [j.id for j in debt_journal]), ('state', '=', 'open')])
-
+        orders = self.env['pos.order'].search([('partner_id', 'in', self.ids)])
+        statements_ids = [st.id for o in orders for st in o.statement_ids]
+        statements = self.env['account.bank.statement.line'].search([('journal_id', 'in', [j.id for j in debt_journal]),
+                                                                     ('state', '=', 'open'),
+                                                                     ('id', 'in', statements_ids)])
```
yelizariev: Instead of
                                                                     ('id', 'in', statements_ids)])

You might use something like ``(pos_order_id.partner_id', 'in', self.ids)``?

yelizariev (changes requested):

pos_debt_notebook/models.py
```diff
@@ -31,19 +31,26 @@ def _compute_debt(self):
...
 
         if statements:
             self._cr.execute(
-                """SELECT l.partner_id, SUM(l.amount)
-                FROM account_bank_statement_line l
-                WHERE l.statement_id IN %s AND l.partner_id IN %s
-                GROUP BY l.partner_id
+                """SELECT po.partner_id, SUM(absl.amount)
+                FROM account_bank_statement_line AS absl
+                    LEFT JOIN pos_order po ON (absl.pos_statement_id=po.id)
```
yelizariev, line 43: po.id and absl.pos_statement_id looks like ids in didfferent tables

yelizariev (changes requested):

pos_debt_notebook/models.py
yelizariev: ok, pos_statement_id is pos.order

pos_debt_notebook/doc/changelog.rst
```diff
@@ -1,3 +1,7 @@
+`4.0.3`
+-------
+- **FIX:** Wrong calculation of partner debts
```
yelizariev: clarify how it worked before

yelizariev (approved):

ilmir-k (approved):

ilmir-k:
@yelizariev it can be merged

PR #745, opened 2018-08-29
:ambulance: Product quantities representation according to the set lo…

KolushovAlexandr:
…cation


yelizariev:
There is more simple way at least in another version https://github.com/it-projects-llc/pos-addons/commit/b83570361569d6072cc5c827bbdc745dd186f315#diff-98128c6f5bc800511052fa25a4bfbbfe

pos_product_available/static/src/js/pos.js
```diff
@@ -3,34 +3,25 @@ odoo.define('pos_product_available.PosModel', function(require){
...
-                        _.extend(self.db.get_product_by_id(prod.id), prod);
+            var loaded = PosModelSuper.load_server_data.call(this);
+
+            return loaded.then(function(){
+                return new Model('product.product').query(['qty_available']).
+                    filter([['sale_ok','=',true],['available_in_pos','=',true]]).
+                    context({'location': self.config.stock_location_id[0]}).all().then(function(data){
+                        _.each(data, function(pr){
+                            self.db.get_product_by_id(pr.id).qty_available = pr.qty_available;
+                        });
```
yelizariev: When pos_cache is not installed you make two requests instead of one. You need to update exising "model" instead

pos_product_available/static/src/js/pos.js
```diff
@@ -3,34 +3,49 @@ odoo.define('pos_product_available.PosModel', function(require){
...
+                };
+                return loaded;
             }
-            this.ready.then(function () {
-                if (self.product_quantities) {
-                    _.each(self.product_quantities, function(prod){
-                        _.extend(self.db.get_product_by_id(prod.id), prod);
+
+            return loaded.then(function(){
+                return new Model('product.product').query(['qty_available']).
```
yelizariev: You make second call when pos_cache is not installed. The data you request here are already loaded

yelizariev (changes requested):

pos_product_available/static/src/js/pos.js
KolushovAlexandr: @yelizariev return on the 37 line prevents it

yelizariev (approved):

pos_product_available/__init__.py
```diff
@@ -0,0 +1,2 @@
+# -*- coding: utf-8 -*-
+from . import models
```
ilmir-k: @KolushovAlexandr no need to import ``models``

ilmir-k (approved):

yelizariev:
Travis needs to be fixed

PR #757, opened 2018-09-19
:heart_eyes: show popup if a table does not exist

GabbasovDinar:
Step to reproduce: 

1. Open POS1 on device 1
2. Open POS2 on device 2
3. in POS1, use edit mode to create a new table, then save the table
4. POS2 has a Popup message

### DIFF
--- a/pos_multi_session/doc/changelog.rst
+++ b/pos_multi_session/doc/changelog.rst
@@ -1,130 +1,134 @@
+`4.2.3`
+-------
+
+**Improvement:** Code refactoring to support the "Sync restaurant orders" module
 
 `4.2.2`
 -------
 
-- **FIX:** Don't make automatic scrolling of the current order on changing other orders
+**Fix:** Don't make automatic scrolling of the current order on changing other orders
 
 `4.2.1`
 -------
 
-- **FIX:** Error related to longpoll updates of paid orders
-- **FIX:** Cannot read property then of undefined or request_sync_all is undefined
+**Fix:** Error related to longpoll updates of paid orders
+**Fix:** Cannot read property then of undefined or request_sync_all is undefined
 
 `4.2.0`
 -------
-- **FIX:** 'order_line.node' is undefined
-- **NEW:** Added synchronization among the same POS
-- **IMP:** Orderlines added to existing orders in offline mode won't be lost after reestablishing a connection
-- **IMP:** Warning message "No connection to the server..." is automatically closing after connection is reestablished
+**Fix:** 'order_line.node' is undefined
+**New:** Added synchronization among the same POS
+**Improvement:** Orderlines added to existing orders in offline mode won't be lost after reestablishing a connection
+**Improvement:** Warning message "No connection to the server..." is automatically closing after connection is reestablished
 
 `4.1.1`
 -------
-- **IMP:** Multi-Session methods renamings
+**Improvement:** Multi-Session methods renamings
 
 `4.1.0`
 -------
-- **NEW:** Field "Active" to enable/disable synchronization for POSes. The multi-session parameter is required now to provide compatibility with "POS Multi Session Menu" module and other modules that put common settings in "POS Multi Session"
+**New:** Field "Active" to enable/disable synchronization for POSes. The multi-session parameter is required now to provide compatibility with "POS Multi Session Menu" module and other modules that put common settings in "POS Multi Session"
 
 `4.0.6`
 -------
-- **FIX:** Incompatibility with pos_order_cancel_restaurant: it showed *Synchronization error* on removing order
+**Fix:** Incompatibility with pos_order_cancel_restaurant: it showed *Synchronization error* on removing order
 
 `4.0.5`
 -------
-- **IMP:** Improved performance
+**Improvement:** Improved performance
 
 `4.0.4`
 -------
-- **FIX:** Remove some orders after revision error
+**Fix:** Remove some orders after revision error
 
 `4.0.3`
 -------
-- **IMP:** All data are updated during POS loading
-- **IMP:** Improved orderline updating speed after synchronization with server
+**Improvement:** All data are updated during POS loading
+**Improvement:** Improved orderline updating speed after synchronization with server
 
 `4.0.2`
 -------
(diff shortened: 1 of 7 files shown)


pos_multi_session/static/src/js/pos_multi_session.js
```diff
@@ -335,28 +335,8 @@ odoo.define('pos_multi_session', function(require){
             var order = new models.Order({}, options);
             return order;
         },
-        ms_update_order: function(order, data){
+        ms_update_exist_order: function(order, data) {
```
yelizariev: existing

pos_multi_session_restaurant/static/src/js/pos_multi_session_restaurant.js
```diff
@@ -95,13 +95,25 @@ odoo.define('pos_multi_session_restaurant', function(require){
...
+            }
+            if (table) {
+                var order = PosModelSuper.prototype.ms_create_order.apply(this, arguments);
                 order.table = self.tables_by_id[data.table_id];
                 order.customer_count = data.customer_count;
+                return order;
             }
-            return order;
+            this.gui.show_popup('confirm',{
+                'title': _t('Not found table'),
```
yelizariev: Got unknown table

yelizariev (approved):

yelizariev:
My comments are fixed :+1:

Ramil-Mukhametzyanov (changes requested):
(-) there is no toolbar to manage the configuration of the tables for demo user
(+) after the modification of the table configuration by the administrator and entering a table, a popup window appears for the demo user and the user can see the changes
(-) after the second modification, a pop-up window for admin appears
(+) but the latest changes for the administrator are still there
(-) demo user can operate on a deleted table
(-) the initialization of tables stops in the "Sync Order" process
The error in the browser console:
point_of_sale.assets.js:299 Uncaught TypeError: Cannot read property 'show' of undefined
	at Class.show_popup (point_of_sale.assets.js:299)
	at child.ms_create_order (point_of_sale.assets.js:661)
	at child.ms_update_order (point_of_sale.assets.js:603)
	at child.ms_update_order (point_of_sale.assets.js:664)
	at point_of_sale.assets.js:594
	at Function._.each._.forEach (web.assets_common.js:12)
	at child.updates_from_server (point_of_sale.assets.js:594)
	at child.updates_from_server (point_of_sale.assets.js:663)
	at Object.<anonymous> (point_of_sale.assets.js:650)
	at fire (web.assets_common.js:541)

Tested in pos-addons 757 runbot with updated and explicit modules

GabbasovDinar:
@Ramil-Mukhametzyanov you can check the PR https://github.com/odoo/odoo/pull/27924 I think it is related

Ramil-Mukhametzyanov:
> @Ramil-Mukhametzyanov you can check the PR [odoo/odoo#27924](https://github.com/odoo/odoo/pull/27924) I think it is related

The Odoo PR [odoo/odoo#27924] is closed, but nothing has changed.

Ramil-Mukhametzyanov (changes requested):
The last changes didn't pass a test with two users admin and demo:
1. If demo user deletes a table and admin clicks on this table, pop-up "unknown table" appears for demo user, not for admin. ( it's the same crosswise)
2. The 'unknown' table displayed on the floor plan. (if we can detect it as unknown we can hide it after the click)
3. Deleting the first table causes continuous loop of errors. (orders are transfered on the first table, but not on the first free table.)

A table must be protected from deleting if it is only one table on the floor.

GabbasovDinar:
> A table must be protected from deleting if it is only one table on the floor.

This question does not apply to this PR. In addition, you can always turn off the table edit setting for the user.

Ramil-Mukhametzyanov (changes requested):
Almost perfect, but one thing should be fixed.

Steps:
1. Delete all the tables by an user
2. Click on a table by another user
Result:
The error 'transfer_table is undefined' in the first user's POS session.

It is necessary to protect the last table from deleting.

Ramil-Mukhametzyanov (approved):
:+1:

yelizariev (changes requested):
No changelog available

pos_multi_session_restaurant/doc/changelog.rst
```diff
@@ -1,3 +1,7 @@
+`3.2.5`
+-------
+- **IMP:** Synchronize the process of deleting or adding tables
```
yelizariev: Use new tags for changelog

pos_multi_session/doc/changelog.rst
```diff
@@ -1,3 +1,7 @@
+`4.2.3`
+-------
+
+- **IMP:** Code refactoring to support the "Sync restaurant orders" module
```
yelizariev: Use new tags

yelizariev (changes requested):
Make just 3 commits for each update: 
*  pos_debt_notebook/doc/changelog.rst (lint)
* pos_multi_session
* pos_multi_session_restaurant

pos_multi_session/static/src/js/pos_multi_session.js
```diff
@@ -334,37 +334,16 @@ odoo.define('pos_multi_session', function(require){
...
             }
         },
-        ms_create_order: function(options){
+        ms_create_order: function(options) {
+            var d = $.Deferred();
             options = _.extend({pos: this}, options || {});
             var order = new models.Order({}, options);
-            return order;
+            d.resolve(order);
+            return d;
```
yelizariev: Doubtful solution

yelizariev (changes requested):

pos_multi_session/doc/changelog.rst
```diff
@@ -1,130 +1,134 @@
+`4.2.3`
+-------
+
+**Improvement:** Code refactoring to support the "Sync restaurant orders" module
```
yelizariev: I don't see why this refactoring is needed

GabbasovDinar: @yelizariev Code refactoring to support **removing/adding tables** for the "Sync restaurant orders" module. Is it correctly?

PR #766, opened 2018-09-26
pos_mobile: fix styles

GabbasovDinar:


yelizariev:
:rainbow: has to be used when *works in the same way, but **code** is better organized, refactored, commented or cleaned up*, which is not the case. Please fix

yelizariev:
Also, please explain in commit message or probably in a code why those updates speed up clicking

GabbasovDinar:
@yelizariev Each time a function is called, the click event is recorded without removing the previous event. When you click the menu button, the event was generated many times, which led to a hang

yelizariev:
@GabbasovDinar Can we just subscribe handler only once?

GabbasovDinar:
@yelizariev the error was caused by incorrect selector definition

yelizariev (dismissed):

yelizariev (approved):

PR #776, opened 2018-10-05
10.0 pos debt notebook generate invoices Updates

KolushovAlexandr:


pos_debt_notebook/models.py
```diff
@@ -179,6 +180,25 @@ def create_from_ui(self, partner):
             del partner['debt_limit']
         return super(ResPartner, self).create_from_ui(partner)
 
+    @api.multi
+    @api.depends('journal_debt', 'journal_credit_balance')
```
yelizariev: Those fields are computed in this method and cannot be used in depends

pos_debt_notebook/models.py
```diff
@@ -179,6 +180,25 @@ def create_from_ui(self, partner):
...
+            domain,
+            fields,
+            'partner_id')
+
+        res_index = dict((id, {'balance': 0}) for id in self.ids)
+
+        for data in res:
+            res_index[data['partner_id'][0]] = data
+        for r in self:
+            r['journal_debt'] = -res_index[r.id]['balance']
```
yelizariev: that it's bracketed and not dot-notation (r.journal_debt)?

pos_debt_notebook/models.py
```diff
@@ -151,6 +150,8 @@ def debt_history(self, limit=0):
     ])
     report_pos_debt_ids = fields.One2many('pos.credit.update', 'partner_id',
                                           help='Technical field for proper recomputations of computed fields')
+    journal_debt = fields.Float(string='Debt per Journal')
+    journal_credit_balance = fields.Float(string='Credit Balance per Journal')
```
yelizariev: You store values in DB that depends on journal and recomputed each time. It doesn't make sense

yelizariev (changes requested):
I'd suggest remove new fields, add return data in _compute_partner_journal_debt

pos_debt_notebook/models.py
KolushovAlexandr: @yelizariev If I make the field computed then it will be recalculated when it is not necessary, also it won't have arguments for that. The field was recalculated, and only then I use its value

yelizariev (approved):
LGTM

PR #778, opened 2018-10-10
:ambulance: pos_invoice diff acc fix

KolushovAlexandr:


yelizariev (approved):

ilmir-k (changes requested):
@KolushovAlexandr We need to consider the case with empty `writeoff_acc_id`. If we have not set ``writeoff_acc_id`` but pay more than invoice amount in POS we could show a warning message to warn that there is no account to put the difference.

yelizariev:
:bath: is only for development. Use :rose: instead

ilmir-k:
> bath is only for development. Use rose instead

Ok

ilmir-k (approved):

PR #783, opened 2018-10-11
:heart_eyes: don't use the 'change' trigger to save to the local storage

GabbasovDinar:
If an order is changed every time this PR fixes an issue with rerendering of the product list. It's related to new changes in Odoo 11.0 https://github.com/odoo/odoo/blame/11.0/addons/point_of_sale/static/src/js/screens.js#L846

### DIFF
--- a/pos_multi_session/doc/changelog.rst
+++ b/pos_multi_session/doc/changelog.rst
@@ -1,3 +1,7 @@
+`4.2.1`
+-------
+- **FIX:** Rerendering the Products list after change a current order
+
 `4.2.0`
 -------
 - **FIX:** 'order_line.node' is undefined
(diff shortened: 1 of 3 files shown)


GabbasovDinar:
@yelizariev Can you check it?

yelizariev (changes requested):
Can there be other reason to use ``trigger('change')``?

Anyway, you have to update change log and *Tested on* note in README

GabbasovDinar:
@yelizariev Who can test these changes? 
image

pos_multi_session/doc/changelog.rst
```diff
@@ -1,6 +1,9 @@
-`4.2.2`
+`4.2.3`
 -------
+- **FIX:** Rerendering the Products list after change a current order
```
yelizariev, line 3: What is the problem with rerendering?

yelizariev (changes requested):

pos_multi_session/doc/changelog.rst
GabbasovDinar: @yelizariev  If we have a rerender of the product list every time, then we will get a hang and slow response of the system. 

How can I change the log message?

GabbasovDinar:
I will make cherry-pick of a commit from https://github.com/it-projects-llc/pos-addons/pull/809 after merging it

PR #801, opened 2018-10-22
12.0 product available

GabbasovDinar:


ilmir-k (dismissed):
@GabbasovDinar  ``website`` must be link to apps store now

Ramil-Mukhametzyanov (changes requested):
After selecting the cashier who has the permission to make payment for the out-of-stock order, the user can not set himself the cashier of the POS.
The user of POS should be set as cashier in the next sessions.

Ramil-Mukhametzyanov (approved):
I can change POS cashier manually.
Approved.

yelizariev (approved):

PR #804, opened 2018-10-26
Merge 10.0 in 11.0

Rusllan:
Merging a 10.0 branch with 11.0


yelizariev:
It's better to test pos_debt_notebook and other modules manually here (if they are not covered via tours enough)

PR #807, opened 2018-10-29
12.0 polling

GabbasovDinar:

PR #808, opened 2018-10-29
12.0 history

GabbasovDinar:
related with #807

### DIFF
--- a/pos_orders_history/README.rst
+++ b/pos_orders_history/README.rst
@@ -26,7 +26,7 @@ Maintainers
 -----------
 * `IT-Projects LLC <https://it-projects.info>`__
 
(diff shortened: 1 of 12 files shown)


GabbasovDinar:
related with https://github.com/it-projects-llc/pos-addons/pull/807

Ramil-Mukhametzyanov (commented):
How to check following function described in the module description?
"You are able to activate the displaying of order barcode on payment receipt. It will allow you to search an order from the list by scanning the barcode."
Using the scanner emulator and writing the letters of the order name form a receipt, I did not succeed.
Among Rule Types in the Default Nomenclature form of the POS settings, I could not find any rule related to the receipts. Trying to create the rule I could not find any type related to the receipts.

Ramil-Mukhametzyanov (approved):
I used xdotool emulator and order name with dashes '-' symbols

yelizariev (approved):

PR #810, opened 2018-10-30
10.0 pos orders history

GabbasovDinar:


ilmir-k (approved):

pos_orders_history/models/pos_orders_history_models.py
```diff
@@ -13,7 +13,10 @@ class PosConfig(models.Model):
     _inherit = 'pos.config'
 
     orders_history = fields.Boolean("Orders History", help="Show all orders list in POS", default=True)
-    current_day_orders_only = fields.Boolean("Current Day Orders Only", help="Show current day orders only", default=True)
+
+    load_orders_of_last_n_days = fields.Boolean("Orders of last 'n' days", default=True)
```
yelizariev: It should be False by defaul

pos_orders_history_return/doc/changelog.rst
```diff
@@ -3,6 +3,12 @@
 Updates
 =======
 
+`1.0.1`
+-------
+
+- **FIX:** Product price is taken from original order rather than Sale Price
```
yelizariev: It's not obvious which variant is correct and which is fixed

pos_orders_history/models/pos_orders_history_models.py
yelizariev: load_limit_orders or something like that sounds better IMO

yelizariev (changes requested):

pos_orders_history/models/pos_orders_history_models.py
GabbasovDinar: @yelizariev Why? We need to load all orders for a specified number of days, not for a specified number of orders

GabbasovDinar: @yelizariev I think first of all that **load_limit_orders** means to load a specified number of orders

yelizariev (approved):

PR #812, opened 2018-10-31
:arrow_up::one::two: pos_barcode_sync

GabbasovDinar:
related with https://github.com/it-projects-llc/pos-addons/pull/807


Ramil-Mukhametzyanov (changes requested):
Unable to enter POS. There is a "this.bus is undefined" or "poll_connection is undefined" error.

Ramil-Mukhametzyanov (approved):

PR #816, opened 2018-10-31
11.0

SouyiroZ:
Hi i can help you with modules and updates i got original and personalized code


yelizariev:
We will do it ourselves after https://github.com/it-projects-llc/pos-addons/pull/817

PR #833, opened 2018-11-14
:ambulance: use correct selector class name

GabbasovDinar:


yelizariev (changes requested):
No changelog updates

GabbasovDinar:
@yelizariev This PR is not ready yet

yelizariev (approved):

Ramil-Mukhametzyanov (changes requested):
1. If I click the "Create invoice" button, the message is displayed: "There is no invoiceable line"
2. After payment, there is the error message on longpolling:
"The operation cannot be completed, probably due to the following: - deletion: you may ..."
3. "There is no invoicable line" (:rose: it's also necessary to correct the spelling).
Tested on mobile device.

ilmir-k:
@Ramil-Mukhametzyanov ``There is no invoicable line`` it is built-in behavior and appears if you have the lines in Sale Order which have ``Invoicing Policy = Delivered`` but not delivered yet.

ilmir-k:
``2. After payment, there is the error message on longpolling:
"The operation cannot be completed, probably due to the following: - deletion: you may ..."``

It is due to empty ``Difference Account``. It makes sense to add exception as this made in version 11.0

Ramil-Mukhametzyanov:
> @Ramil-Mukhametzyanov `There is no invoicable line` it is built-in behavior and appears if you have the lines in Sale Order which have `Invoicing Policy = Delivered` but not delivered yet.

@ilmir-k This message appears for non-invoiced orders, and then POS proceeds with payment screen. You can check it.

Ramil-Mukhametzyanov (changes requested):
image
1. The status value "Open" does not fit on the screen.
2. Column cells have offset relative to the header

Ramil-Mukhametzyanov (approved):

Issue #819, opened 2018-11-03
[11][pos_restaurant_base] this is window instead of order


ruben-tonetto:
in print_order_receipt function this is global variable window, instead of order.
so variable widged passed to receipt doesn't contain order data

maybe:

$.proxy( function(){
                    var receipt = QWeb.render('OrderChangeReceipt',{changes:changes, widget:this});
                    printer.print(receipt);
                    return delay(100);
                }, this )

GabbasovDinar:
Hello @ruben-tonetto 

Yes, you are right. We need to use `self ` instead `this`

GabbasovDinar:
@yelizariev This error is present in all versions. I will fix for 10.0 but we need to port to newer versions (11.0 and 12.0)

ruben-tonetto:
Mmm... did you test it?
self is a reference to this, but when this changes to window, self refers to window.
i've tested and seems not working

PS: thanks for your projects

GabbasovDinar:
@ruben-tonetto How did you test it?

I don't have the error

![untitled](https://user-images.githubusercontent.com/15119940/48614091-3efb3800-e9af-11e8-8d9e-10f5eb3bbe7c.png)

ruben-tonetto:
Sorry, you are right!
It works
thanks

Il giorno ven 16 nov 2018 alle ore 10:52 Dinar <notifications@github.com>
ha scritto:

gustavovalverde:
@GabbasovDinar this can be closed

GabbasovDinar:
@yelizariev Please, close the issue

PR #835, opened 2018-11-15
:ambulance: invalid pointer to the context

GabbasovDinar:
close https://github.com/it-projects-llc/pos-addons/issues/819


yelizariev:
It doesn't fix according to https://github.com/it-projects-llc/pos-addons/issues/819#issuecomment-439172213

GabbasovDinar:
@yelizariev 
image

PR #851, opened 2018-11-29
:tada::one::zero: pos_receipt_custom and pos_order_receipt_custom modules

ilmir-k:
:tada::one::zero: pos_receipt_custom and pos_order_receipt_custom modules


GabbasovDinar:
image

GabbasovDinar:
> The Function constructor is eval no-new-func

@yelizariev How can I fix this lint error? I need to use this construction `new Function(['dict'], code);` because variable `code` can be any value. This construction allows compiling the function with arbitrary code. I took this code from odoo https://github.com/odoo/odoo/blob/10.0/addons/web/static/lib/qweb/qweb2.js#L464

yelizariev:
Why don't you call built-in render methods?

GabbasovDinar:
@yelizariev I call built-in render methods here https://github.com/it-projects-llc/pos-addons/blob/a8f62e0734fd8112c69d9aaa0d272a8ce24f19ca/pos_receipt_custom/static/src/js/models.js#L82 but we need to set qweb code for the variable  https://github.com/it-projects-llc/pos-addons/blob/dfa1893aa743d62eaf4d38ca5757fde484c66c9c/pos_receipt_custom/static/src/js/models.js#L81
We can only use the built-in function, but we need to override the function. I believe that it is not advisable to override the method that can lead to errors in generating qwebs for other modules. It is wrong to override the entire framework for the sake of a single module.

Ramil-Mukhametzyanov (approved):

PR #809, opened 2018-10-29
12.0 multi session

GabbasovDinar:
related with https://github.com/it-projects-llc/pos-addons/pull/807

PR #852, opened 2018-11-29
:ambulance: fix rerendering the Products list after change a current order

GabbasovDinar:
Before this fix, if we have a rerender of the product list every time, then we will get a hang and slow response of the system.

P.S. this PR is made using cherry-pick from https://github.com/it-projects-llc/pos-addons/pull/809/commits/fab34c7623d7951278eff5fa663587d9d10c28f6


yelizariev (changes requested):
Version is not updated

yelizariev (approved):
Code looks good

Ramil-Mukhametzyanov (approved):

PR #859, opened 2018-12-10
:book: languages are not supported using esc-pos

Ramil-Mukhametzyanov:


yelizariev:
Merged in https://github.com/it-projects-llc/pos-addons/commit/0000a2a1b0055ffa3af92e15054d9729f632048d

PR #872, opened 2018-12-20
:heart_eyes: compatibility with pos_order_receipt_custom

GabbasovDinar:


Ramil-Mukhametzyanov (changes requested):
Issue related to using one of the sessions without an order printer
Steps:
1. Add item in the session (Admin) with order printer in the Table T10. Print order.
2. Move this order to the Table T3 in a session (Demo) without order printer
3. Print order in the session (Admin).
Result: 
      printed as "Cancelled", not "Transfered".
      lost the information about the previous table.
image

Ramil-Mukhametzyanov (changes requested):
Using sessions with order printers:
If transfer an order from one table to another, a TRANSFER order is printed. It's OK.
BUT Order button from another session still is active .
After click on it, 2 ticket are printed: CANCEL order and ADD order
image

GabbasovDinar:
@Ramil-Mukhametzyanov Can you repeat the errors for the module v10.0?

GabbasovDinar:
@Ramil-Mukhametzyanov Can you reproduce the error without this module?

GabbasovDinar:
I close this PR, because the updates updates will be merged from 10.0 to 11.0, then from 11.0 to 12.0 after merging the https://github.com/it-projects-llc/pos-addons/pull/876

PR #876, opened 2018-12-20
:heart_eyes: compatibility with pos_order_receipt_custom

GabbasovDinar:
Task: https://www.it-projects.info/web#id=1729&view_type=form&model=project.task

After merging this PR we need to create new PR to 11.0 and 12.0 versions

### DIFF
--- a/pos_multi_session/static/src/js/pos_multi_session.js
+++ b/pos_multi_session/static/src/js/pos_multi_session.js
@@ -230,7 +230,7 @@ odoo.define('pos_multi_session', function(require){
                     this.get('selectedOrder').ms_replace_empty_order = true;
                     return;
                 } else if (this.ms_syncing_in_progress){
-                    if (this.get('orders').size() === 0) {
+                    if (this.get('orders').size() === 0 && this.gui.get_current_screen() !== "floors") {
                         this.add_new_order();
                     } else {
                         return this.set({'selectedOrder': this.get('orders').at(index) || this.get('orders').first()});
--- a/pos_multi_session_restaurant/__manifest__.py
+++ b/pos_multi_session_restaurant/__manifest__.py
@@ -10,7 +10,7 @@
     "category": "Point of Sale",
     "live_test_url": 'http://apps.it-projects.info/shop/product/pos-multi-session?version=10.0',
     "images": ['images/s2.png'],
-    "version": "10.0.3.2.5",
+    "version": "10.0.3.2.6",
     "application": False,
 
     "author": "IT-Projects LLC, Ivan Yelizariev",
--- a/pos_multi_session_restaurant/static/src/js/pos_multi_session_restaurant.js
+++ b/pos_multi_session_restaurant/static/src/js/pos_multi_session_restaurant.js
@@ -88,8 +88,8 @@ odoo.define('pos_multi_session_restaurant', function(require){
         add_new_order: function(){
             var self = this;
             PosModelSuper.prototype.add_new_order.apply(this, arguments);
-            if (this.multi_session){
-                var current_order = this.get_order();
+            var current_order = this.get_order();
+            if (this.multi_session && current_order){
                 current_order.new_updates_to_send();
                 current_order.save_to_db();
             }
@@ -183,10 +183,6 @@ odoo.define('pos_multi_session_restaurant', function(require){
                     return ord.uid === data.uid;
                 });
             }
-            if (order && order.table && order.table.id !== data.table_id) {
-                order.transfer = true;
-                order.destroy({'reason': 'abandon'});
-            }
             PosModelSuper.prototype.updates_from_server.apply(this, arguments);
             if ((order && old_order && old_order.uid !== order.uid) || (old_order === null)) {
                 this.set('selectedOrder',old_order);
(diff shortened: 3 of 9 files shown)


Ramil-Mukhametzyanov (changes requested):
In the case of two active multisessions with configured order printers, two TRANSFER orders are printed.
In one session (Admin) do the following:
1. Add item, print -> printed NEW
2. Add item, print -> printed ADD
3. Remove the last item, print -> printed CANCEL
4. Change table -> 
printed TRANSFER (Printer Waiter: Demo User)
and another TRANSFER (Printer Waiter: Administrator)

Ramil-Mukhametzyanov (changes requested):
1. Add item in order in the session of Admin. Print order.
2. Move this order to the another table in the session of Demo
    Result: on TRANSFER ticket Printer waiter: Administrator (expected Demo)
If create an order by Demo, and then transfer it by Administrator, the value of Printer waiter on the TRANSFER ticket will be Demo.

Ramil-Mukhametzyanov (changes requested):
The order button is active for other user after the transfer.

Ramil-Mukhametzyanov (changes requested):
1. Add item by Demo, print -> printed NEW
2. Add item, by Administrator, print ->
     printed NEW (expected ADD)

GabbasovDinar:
@Ramil-Mukhametzyanov I fixed the problems

Ramil-Mukhametzyanov (changes requested):
Lost of synchronization.

Demo:
   1. Select a table.
Admin: 
   2. Open the same table and the same tab
   3. Add an item, print order (click Order)
   4. Move the order to another table
Can see: the table is changed for Demo too.
Demo:
   5. Destroy the order (click '-')
Result: the order is not destroyed in session of Admin

Ramil-Mukhametzyanov (commented):
In the last described case, it is best to print a CANCEL order and delete the order.

pos_multi_session_restaurant/__manifest__.py
```diff
@@ -10,7 +10,7 @@
     "category": "Point of Sale",
     "live_test_url": 'http://apps.it-projects.info/shop/product/pos-multi-session?version=10.0',
     "images": ['images/s2.png'],
-    "version": "10.0.3.2.5",
+    "version": "12.0.3.2.6",
```
Ramil-Mukhametzyanov: Here is a mistake: Odoo version is 10.0, not 12.0

Ramil-Mukhametzyanov (changes requested):
1. Create an order by Demo, print order
    In my example the order 00001-001-0001, Main Floor/T1
2. Find the order by Admin and delete it (destroy by clicking '-')
3. Create new order by Admin, print order
    In my example it was 00002-001-002, Main Floor/T2 (Extra Flandria chicory 1kg)
4. Destroy it by Admin.
5. Click on a table in the Demo's session
    I clicked on the table T3
Result:
    The TRANSFER order is printed
    In my case, it was 00002-001-0002 Main Floor/T2>>>Main Floor/T3 (Extra Flandria chicory 1kg)

It is better to do following oprations after clicking '-' button (if pos_order_receipt_custom module is installed):
1. deletion all the lines (as we click 'delete lines' button)
2. print CANCEL order
3. and the destroy the empty order.

Ramil-Mukhametzyanov (changes requested):
After the last modification, the new order will not appear to another multi-session user.

Ramil-Mukhametzyanov (changes requested):
After completing the above steps, any NEW orders will no longer print.

Ramil-Mukhametzyanov (approved):

pos_multi_session_restaurant/static/src/js/pos_multi_session_restaurant.js
```diff
@@ -183,9 +183,12 @@ odoo.define('pos_multi_session_restaurant', function(require){
...
             }
-            if (order && order.table && order.table.id !== data.table_id) {
-                order.transfer = true;
-                order.destroy({'reason': 'abandon'});
+            if (order && order.table && order.table.id !== data.table_id && message.action !== "remove_order") {
+                // for compatibility with https://www.odoo.com/apps/modules/10.0/pos_order_receipt_custom/ module
+                order.printed_transfer = data.printed_transfer;
+                this.order_to_transfer_to_different_table = order;
+                this.set_table(this.tables_by_id[data.table_id]);
+                this.order_to_transfer_to_different_table = null;
```
yelizariev: Why this cannot be done in the module pos_order_receipt_custom?

yelizariev (changes requested):
Надо из мульти-сессии убрать код, который чисто под pos_order_receipt_custom сделан

pos_multi_session_restaurant/static/src/js/pos_multi_session_restaurant.js
```diff
@@ -205,18 +201,25 @@ odoo.define('pos_multi_session_restaurant', function(require){
                 order.init_locked = false;
             }
         },
+        ms_update_existing_order: function(order, data) {
+            // the order has been transferred to another table
```
yelizariev: Either say "Check that the order has been ..." or move it inside if-block

yelizariev (approved):
Looks better

pos_multi_session/static/src/js/pos_multi_session.js
```diff
@@ -230,7 +230,7 @@ odoo.define('pos_multi_session', function(require){
                     this.get('selectedOrder').ms_replace_empty_order = true;
                     return;
                 } else if (this.ms_syncing_in_progress){
-                    if (this.get('orders').size() === 0) {
+                    if (this.get('orders').size() === 0 && this.gui.get_current_screen() !== "floors") {
```
Ramil-Mukhametzyanov, line 233: We need changelog and manifest updates for pos_multi_session

Ramil-Mukhametzyanov (changes requested):
After the following steps are performed, the an opened order is displayed in the Admin's POS.
Demo:
1    Select a table.
Admin:
2    Open the same table and the same tab
3    Add an item (for example, Boni Oranges 1kg), print order (click Order)
4    Move the order to another table
Can see: the table is changed for Demo too.
Demo:
5   Destroy the order (click '-')
Result: an open order is displayed in the current table of Admin (even if the order is in another table)
The order payment from this table is successful (receipt data is correct).
But if the order has the same products (for example, Boni oranges 1kg), it may be confusing.

It is best to leave the table.

GabbasovDinar:
@Ramil-Mukhametzyanov This problem does not related to this PR

Ramil-Mukhametzyanov (approved):
ok

PR #891, opened 2018-12-28
10.0 pos order receipt custom fix

GabbasovDinar:


Ramil-Mukhametzyanov (changes requested):
In the receipt screen, the Change line is below the barcode

Ramil-Mukhametzyanov (approved):
May skip it for a while

yelizariev (approved):

PR #911, opened 2019-01-25
:ambulance: pos_printer_network: incorrect identify the PosBox url

GabbasovDinar:


Ramil-Mukhametzyanov (approved):
Tested on runbot. It seems that we have found the root of the problem.

yelizariev (approved):
Automatic computation of the url looks suspicious but ok

PR #912, opened 2019-01-28
:zap: New button to add all partners with debts, availability to upda…

KolushovAlexandr:
…te credits via Manual Credit Updates

### DIFF
--- a/pos_debt_notebook/wizard/pos_credit_invoices.py
+++ b/pos_debt_notebook/wizard/pos_credit_invoices.py
@@ -109,7 +109,7 @@ class PosCreditInvoices(models.TransientModel):
                     'update_type': self.update_type,
                 }
                 if self.update_type == 'new_balance':
-                    vals['new_balance'] = line.amount
+                    vals['new_balance'] = self.new_balance
                 else:
                     vals['balance'] = - line.amount
                 credit_update = self.env['pos.credit.update'].create(vals)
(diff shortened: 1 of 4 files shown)


pos_debt_notebook/wizard/pos_credit_invoices.py
```diff
@@ -22,11 +22,12 @@ class PosCreditInvoices(models.TransientModel):
     product_id = fields.Many2one(
         'product.product',
         'Credit Product',
-        required="True",
         help="This product will be used on creating invoices."
     )
+    use_invoices = fields.Boolean('Via Invoices', default=True,
+                                  help='Otherwise it will be executed via "Manual Credit Updates"')
```
yelizariev: Consider to use Selection instead.

pos_debt_notebook/wizard/pos_credit_invoices_views.xml
```diff
@@ -13,11 +13,16 @@
...
 
                         <field name="journal_id"/>
-                        <field name="product_id"/>
+                        <field name="product_id" attrs="{'required':[('use_invoices', '=', True)],
+                                                         'invisible':[('use_invoices', '=', False)]}"/>
                     </group>
                 </group>
+                <button name="add_partners_with_debt" string="Add All Partners"
+                        type="object" class="btn-primary"
+                        attrs="{'invisible':[('update_type', '=', 'balance_update')]}"/>
```
yelizariev, line 25: Why just don't make a checkbox? Handling a thousand partners via ui could slow the workflow down

yelizariev (changes requested):

pos_debt_notebook/wizard/pos_credit_invoices_views.xml
KolushovAlexandr: As it appeared, the implemented method is more convenient, when you need to specify all partners except, for example, two ones.

PR #914, opened 2019-01-28
[ADD] Allow price to be editable depending on product flag

ssaid:


yelizariev (commented):
Looks interesting. What do you think about making this access global, rather than per product? Do you have scenarios when some products should be modifiable and some should not?

ssaid:
@yelizariev Thank you by your comment, yes we have certain situations that cannot be controlled with global access.

PR #930, opened 2019-02-26
:heart_eyes: disable mobile mode in POS

GabbasovDinar:


Ramil-Mukhametzyanov (approved):

pos_mobile/models/pos_config_model.py
```diff
@@ -0,0 +1,10 @@
+# Copyright 2019 Dinar Gabbasov <https://it-projects.info/team/GabbasovDinar>
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
+
+from odoo import models, fields
+
+
+class PosConfig(models.Model):
+    _inherit = 'pos.config'
+
+    mobile_version_is_active = fields.Boolean(string="Mobile version is Active", default=True)
```
ilmir-k: @GabbasovDinar 
I think it's better to name it ``Mobile version`` instead

yelizariev: ``allow_mobile`` + Update description so it's easier to understand Meaning (It's allow mode, not a force mode)

yelizariev (changes requested):
Use proper variable name and clarify docs

pos_mobile/controllers/main.py
```diff
@@ -0,0 +1,29 @@
...
+        if not pos_sessions:
+            return werkzeug.utils.redirect('/web#action=point_of_sale.action_client_pos_menu')
+        pos_sessions.login()
+        if pos_sessions.config_id.mobile_version_is_active:
+            session_info = request.env['ir.http'].session_info()
+            session_info['is_mobile'] = True
+            context = {
+                'session_info': json.dumps(session_info),
+            }
+            return request.render('point_of_sale.index', qcontext=context)
```
yelizariev: It's better call super and update context in response
See https://github.com/odoo/odoo/blob/10.0/addons/website_portal_sale/controllers/main.py#L14

pos_mobile/controllers/main.py
```diff
@@ -0,0 +1,29 @@
...
+            ('state', '=', 'opened'),
+            ('user_id', '=', request.session.uid),
+            ('rescue', '=', False)])
+
+        if not pos_sessions:
+            return werkzeug.utils.redirect('/web#action=point_of_sale.action_client_pos_menu')
+        pos_sessions.login()
+        if pos_sessions.config_id.mobile_version_is_active:
+            session_info = request.env['ir.http'].session_info()
+            session_info['is_mobile'] = True
```
yelizariev: If ``mobile_version_is_active``, then it doesn't mean that system *is mobile*, it's only *allow mobile*

pos_mobile/views/pos_mobile_template.xml
```diff
@@ -50,7 +50,7 @@ var MobileCheck = function MobileCheck() {
 })()
 
 // exclude pos_mobile from module list if the url doesn't have parameter "m" in agrs (e.g. /pos/web?m=1 )
```
yelizariev: This comment is obsolete and need to be updated

pos_mobile/views/pos_mobile_view.xml
```diff
@@ -0,0 +1,23 @@
...
+            <field name="arch" type="xml">
+                <xpath expr="//field[@name='iface_big_scrollbars']/../.." position="after">
+                    <div class="col-xs-12 col-md-6 o_setting_box" >
+                        <div class="o_setting_left_pane">
+                            <field name="mobile_version_is_active"/>
+                        </div>
+                        <div class="o_setting_right_pane">
+                            <label for="mobile_version_is_active"/>
+                            <div class="text-muted">
+                                Disable or Enable Mobile Version
```
yelizariev: You cannot *Enable* Mobile via this option. You can only *Allow*

yelizariev (changes requested):

yelizariev:
More flexible solution might be this:

     auto_mobile = fields.Boolean ("Auto Mobile View", help="Switch to Mobile  view automatically depending on device", default=True)

pos_mobile/controllers/main.py
```diff
@@ -0,0 +1,24 @@
...
+import json
+
+from odoo import http
+from odoo.http import request
+from odoo.addons.point_of_sale.controllers.main import PosController
+
+
+class ControllerPos(PosController):
+
+    @http.route('/pos/web', type='http', auth='user')
```
yelizariev: ``@http.route()`` is enough

pos_mobile/static/src/js/mobile.js
```diff
@@ -6,7 +6,7 @@ odoo.define('pos_mobile.mobile', function (require) {
     var PosModelSuper = models.PosModel.prototype;
     models.PosModel = models.PosModel.extend({
         initialize: function (session, attributes) {
-            this.is_mobile = odoo.is_mobile;
+            this.is_mobile = odoo.is_mobile && odoo.session_info.allow_mobile;
```
yelizariev: It evaluate to False when auto_mobile is False, but ``m=1``

yelizariev (changes requested):

yelizariev (approved):

yelizariev:
Change tag to :zap: (new feature) please

PR #931, opened 2019-02-26
:heart_eyes: load order after scan barcode

GabbasovDinar:


ilmir-k (changes requested):
@GabbasovDinar update the version please

yelizariev:
Are you sure that both updates are just :heart_eyes:  and not  :ambulance: or :zap: (depending on whether is it supposed feature or not)

:heart_eyes: = 
before the commit it worked in expected way, but now it works better (quicker, nicer)

KolushovAlexandr (approved):

ilmir-k (approved):

PR #932, opened 2019-02-28
:book: use the port according to nginx configuration

ilmir-k:
:book: use the port according to nginx configuration


TestMergeBot (commented):
No features needs to be tested

yelizariev (approved):

PR #936, opened 2019-03-05
:heart_eyes: No need to specify a Receipt Printer in config for using…

GabbasovDinar:
… Kitchen Printers

### DIFF
--- a/pos_printer_network/README.rst
+++ b/pos_printer_network/README.rst
@@ -6,8 +6,6 @@ Print POS orders and receipts by using a network printer.
 
 By default, odoo supports usb printers only for POS order printing. The module makes it possible to print POS orders and receipts on network printers.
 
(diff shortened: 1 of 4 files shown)


Ramil-Mukhametzyanov (changes requested):
Even if the USB printer is switched on, its status is shown as 'offline'
usb printer

Ramil-Mukhametzyanov (changes requested):
The error 'self.devices_status is undefined' appears after turning off the USB printer.
Steps to reproduce:
1. usb printer is on
2. enter pos
3. switch off  usb printer
4. reload pos, click on network printer icon
5. reload pos, click on network printer icon again

Ramil-Mukhametzyanov (approved):

yelizariev (approved):
Consider splitting commit into two as there are two updates and not just one

PR #940, opened 2019-03-13
:tada::one::zero: pos_order_print_check module

ilmir-k:
:tada::one::zero: pos_order_print_check module


TestMergeBot (commented):
Some features needs to be tested

pos_order_print_check/doc/changelog.rst
```diff
@@ -0,0 +1,4 @@
+`1.0.0`
+-------
+
+- Init version
```
TestMergeBot, line 4: Someone has to test it

Ramil-Mukhametzyanov (approved):

pos_order_print_check/static/src/js/multiprint.js
```diff
@@ -0,0 +1,36 @@
...
+                var printer = printers.find(function(p){
+                    return p.ip === self.config.proxy_ip;
+                });
+                if (printer && printer.status === "online") {
+                    done.resolve();
+                } else {
+                    done.reject();
+                }
+            } else {
+                $.ajax({
```
KolushovAlexandr, line 21: Not sure but probably you can just return ajax request in 'else' to avoid duplicating resolving/rejecting for 'done'. Because 'done' and 'ajax' behave the same in that case. And assigning/returning of the 'done' may be put into the 'if' then.

KolushovAlexandr (approved):

PR #945, opened 2019-03-14
:ambulance: error of using different companies within one multi-session

GabbasovDinar:


TestMergeBot (commented):
Some features needs to be tested

pos_multi_session/doc/changelog.rst
```diff
@@ -1,3 +1,8 @@
+`4.2.6`
+-------
+
+**Fix:** The error associated with the use of different companies within the same multi-session
+
```
TestMergeBot: Someone has to test it

pos_multi_session/doc/changelog.rst
```diff
@@ -1,3 +1,8 @@
+`4.2.6`
+-------
+
+**Fix:** The error associated with the use of different companies within the same multi-session
```
yelizariev: Don't use "associated". You can say "error on using different companies..."

yelizariev (changes requested):

Ramil-Mukhametzyanov (changes requested):
If log in as Demo and change the company, I will see the corresponding multisession POSes.
But if I enter as Admin, I will see every POS no matter which company I choose.

Ramil-Mukhametzyanov (approved):
In the last comment is the Odoo 11.0 feature.

yelizariev (approved):

PR #947, opened 2019-03-15
:ambulance: revision error did not remove deleted orders

KolushovAlexandr:


pos_multi_session_sync/doc/changelog.rst
```diff
@@ -1,3 +1,7 @@
+`1.0.5`
+-------
+- **Fix:** Revision Error did not remove deleted orders
```
yelizariev: too technical. Describe what was the problem from user point of view (unsynced order in some cases)

yelizariev (changes requested):

PR #951, opened 2019-03-20
10.0 pos order cancel integaration with pos pin

KolushovAlexandr:


pos_pin/doc/changelog.rst
```diff
@@ -3,6 +3,11 @@
 Updates
 =======
 
+`1.2.0`
+-------
+
+- **NEW:** Availability not to change cashier after entering pin
```
yelizariev: Ability?

pos_pin/static/src/js/pos.js
```diff
@@ -38,20 +43,42 @@ odoo.define('pos_pin.pos', function (require) {
...
                     });
                 } else {
-                    return self.set_and_render_cashier(user);
+                    return def_pass.resolve();
                 }
+            }).fail(function(res){
+                return def_pass.reject(res);
             });
+
+            return def_pass.done(function(res){
```
yelizariev: It seems that you don't need ``def_pass``. 


    Try def_list.done(...the same here, but without def_pass..).done(....check_then_set_and_render_cashier....)

pos_pin/static/src/js/pos.js
```diff
@@ -38,20 +43,42 @@ odoo.define('pos_pin.pos', function (require) {
...
+                return self.check_then_set_and_render_cashier(options, res);
+            }).fail(function(res){
+                return res;
+            });
+        },
+
+        check_then_set_and_render_cashier: function(options, user) {
+            if (options.do_not_change_cashier){
+                return user;
+            }
```
yelizariev, line 65: I don't see a reason to have this in separate method

yelizariev (changes requested):

gustavovalverde (changes requested):
Just as advanced order notes works. This one should have an option to Ask for the manager just when the order was sent to the kitchen.

I think you'll have to make an extra module for that functionality, as it will depend on pos_restaurant; making this module for restaurants and not as a generic one <-- Your call.

gustavovalverde (approved):

pos_order_cancel_restaurant/models/models.py
```diff
@@ -13,6 +13,7 @@ class PosConfig(models.Model):
     save_canceled_kitchen_orders_only = fields.Boolean(string="Save Canceled / Refunded kitchen orders only",
                                                        default=False, help="Save canceled / refunded orders in Backend")
     send_removed_lines_to_kitchen = fields.Boolean(string="Send removed line to Kitchen", default=True, help="If unchecked, the orderline will not be printed repeatly after its removing")
+    ask_managers_pin_kitchen_orders_only = fields.Boolean(string="Ask Manager For Kitchen orders only", default=False, help="Only Managers are allowed to cancel kitchen order lines")
```
Ramil-Mukhametzyanov, line 16: This option does not work properly.
The 'Select User' popup is displayed, if only an order is created, all lines are deleted and then the order deleting button [-] is clicked.

Ramil-Mukhametzyanov (changes requested):

pos_order_cancel_restaurant/models/models.py
KolushovAlexandr: Did you send the order to the kitchen? It should work only that way

Ramil-Mukhametzyanov: No, I did not.
I have tested trying to delete an order lines that has been sent to the kitchen. It works.

Ramil-Mukhametzyanov (approved):

yelizariev (approved):

PR #964, opened 2019-03-31
11.0 719922f491313952462a767428a701567295d61f

KolushovAlexandr:


Ramil-Mukhametzyanov (changes requested):
Using the multi_session option "One Waiter per Table" clicking on a table will result in the error "cashier is undefined"

pos_keyboard/__manifest__.py
```diff
@@ -7,7 +7,7 @@
     'images': ['images/keyboard.png'],
     'category': 'Point Of Sale',
     'license': 'LGPL-3',
-    'version': '10.0.1.0.2',
+    'version': '10.0.1.1.0',
```
Ramil-Mukhametzyanov, line 10: The module could bot be tested on runbot due to a wrong version number.

Ramil-Mukhametzyanov: odoo 11.0

Ramil-Mukhametzyanov (changes requested):
"Ask manager" option: when the demo (POS user) tries to delete the order, there is the same type of error.

Ramil-Mukhametzyanov (changes requested):
pos_partner_sync:
If the customer list screen is not open, new barcode will not be applied to a partner in another POS.

pos_keyboard/__manifest__.py
KolushovAlexandr: @Ramil-Mukhametzyanov, it's not ported yet, so it's not installable. 
https://github.com/it-projects-llc/pos-addons/pull/964/files/9ca66a6b2dbfbc3ababdda04af9760f2008a1092#diff-036e3bb5eeff71c6285f911b70b19b8dR17
And I have the question, should i do it in this pr or in the another?

Ramil-Mukhametzyanov: OK, you should not port it in this pr.

Ramil-Mukhametzyanov (changes requested):
Error after click 'cancel' in the lock screen.
Steps to reproduce:
1. Set PIN for Admin. Turn on 'Ask Manager' option in the Demo's POS.
2. Enter POS as Demo, create an order, add an item and click to remove the item.
3. Select Admin, then click Cancel
Result: 'cancel_function is undefined' and, in addition, the item has been is deleted.

Ramil-Mukhametzyanov (approved):

PR #966, opened 2019-04-04
:ambulance: client lines updating on the clientlistscreen

KolushovAlexandr:


Ramil-Mukhametzyanov (changes requested):
When I try to edit a company's name from the POS customer list screen, an 'partner is not defined' error occurs.

pos_multi_session/static/src/js/pos_multi_session.js
```diff
@@ -120,7 +120,7 @@ odoo.define('pos_multi_session', function(require){
                 this.message_ID = 0;
             }
             this.multi_session = false;
-            this.ms_syncing_in_progress = false;
+            this.ms_syncing_in_pzogress = false;
```
Ramil-Mukhametzyanov: Is it intentional change?

Ramil-Mukhametzyanov (approved):
perfect :+1:

pos_partner_sync/static/src/js/pos_partner_sync.js
```diff
@@ -55,6 +68,11 @@ odoo.define('pos_partner_sync.pos', function (require) {
             : [ids];
             new Model(model_name).call("read", [ids, fields], {}, {'shadow': true}).then(function(partners) {
                 // check if the partners we got were real updates
+                // we add this trick with get_partner_write_date to be able to process several updates within the second
+                // it is restricted by built-in behavior in add_partners function
+                self.db.partner_write_date = 0;
+                // returns default value "1970-01-01 00:00:00"
+                self.db.get_partner_write_date();
```
yelizariev: Senseless

yelizariev (changes requested):
load_new_partners no need to be redefined. Just use new method, e.g. load_new_partners_force_update.

yelizariev:
load_new_partners_force_update must have comment about what is difference with load_new_partner

pos_partner_sync/static/src/js/pos_partner_sync.js
```diff
@@ -53,10 +53,9 @@ odoo.define('pos_partner_sync.pos', function (require) {
             }
         },
 
-        load_new_partners: function(ids){
-            if (!ids) {
-                return PosModelSuper.prototype.load_new_partners.apply(this, arguments);
-            }
+        load_new_partners_force_update: function(ids) {
+            // quite similar to load_new_partners but loads only required partners and do it forcibly (see the comment below)
+            ids = ids || this.db.partner_sorted;
```
yelizariev: Does that mean, that  it updates ALL partners (even without any updates), when you call method without parameters? Is it really necessary?

KolushovAlexandr: @yelizariev  I made it work by analogy with `load_new_partners` method. So as it's written in the name of the function it is the same as `load_new_partners` but do it forcibly. It is not necessary, i can change it to a check for `ids.length` and return rejected defered object if `ids.length === 0`

yelizariev: Yes, make it impossible to use without ids

yelizariev (approved):

PR #973, opened 2019-04-10
:zap: For different POS categories available are different Product Notes

GabbasovDinar:


Ramil-Mukhametzyanov (changes requested):
When a product with non-specified category is selected, the "category_ids is undefined" error occurs after clicking Note button.

Ramil-Mukhametzyanov (approved):

pos_order_note/static/src/js/order_note.js
```diff
@@ -15,7 +15,10 @@ odoo.define('pos_order_note', function (require) {
 
     models.load_models({
         model: 'pos.product_notes',
-        fields: ['name', 'sequence'],
+        fields: ['name', 'sequence', 'pos_category_ids'],
+        condition: function(self){
+            return self.config.module_pos_restaurant;
```
yelizariev, line 21: Add a comment to code about what does that mean

pos_order_note/doc/changelog.rst
```diff
@@ -1,3 +1,8 @@
+`1.3.0`
+-------
+
+- **NEW:** For different POS categories available are different Product Notes
```
yelizariev: image

yelizariev (changes requested):

pos_order_note/doc/changelog.rst
```diff
@@ -1,3 +1,8 @@
+`1.3.0`
+-------
+
+- **NEW:** Each POS category has its own set of notes for POS products.
```
yelizariev: Sorry, but it's actually opposite -- each note can be applied for specific set of categories only

pos_order_note/models/models.py
```diff
@@ -12,3 +12,6 @@ class PosProductNotes(models.Model):
 
     sequence = fields.Integer(string="Sequence")
     name = fields.Char(string="Note")
+    pos_category_ids = fields.Many2many('pos.category', string='Point of Sale Categories',
+                                        help='The note will be available for this group of POS categories.'
```
yelizariev: I think space at the end is needed. Otherwise it's ``The note will be available for this group of POS categories.Leave the field empty so that the note is available for all POS categories.``

bealdav:
```suggestion
                                        help='The note will be available for this group of POS categories. '
```

PR #974, opened 2019-04-10
:ambulance: Error during installation in multi-company mode

GabbasovDinar:
We have the `The chosen company is not in the allowed companies for this user` error if we try to install the pos_multi_session module with a demo data in the Multi-Company mod. This PR fix the issue (when creating a new user, the current company is assigned by default).

### DIFF
--- a/pos_multi_session/demo/demo.xml
+++ b/pos_multi_session/demo/demo.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <!-- Copyright 2016 Ilyas Rakhimkulov
-     Copyright 2016,2018 Dinar Gabbasov <https://it-projects.info/team/GabbasovDinar>
+     Copyright 2016,2018-2019 Dinar Gabbasov <https://it-projects.info/team/GabbasovDinar>
      Copyright 2016 Ivan Yelizariev <https://it-projects.info/team/yelizariev>
      Copyright 2017 Kolushov Alexandr <https://it-projects.info/team/KolushovAlexandr>
      License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl.html). -->
@@ -26,7 +26,6 @@
 
         <record id="partner_demo2" model="res.partner">
             <field name="name">Demo User2</field>
-            <field name="company_id" ref="base.main_company"/>
             <field name="customer" eval="False"/>
             <field name="email">demo2@yourcompany.example.com</field>
             <field name="street">Avenue des Dessus-de-Lives2, 2</field>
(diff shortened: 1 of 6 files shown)


Ramil-Mukhametzyanov (changes requested):
Using multi-company.
After installing pos_multi_session module, the company in all POS configurations will set to current company.
Even after restoring the initial company in the POS configuration, if the demo user chooses the same company that was selected before the admin logout, he can view these POSs. This change has not been applied.

pos_multi_session/demo/demo.xml
```diff
@@ -17,16 +17,12 @@
             <field name="name">POS2</field>
             <field name="multi_session_id" ref="demo_multi_session"/>
         </record>
-        <record id="demo_multi_pos_config_three" model="pos.config">
-            <field name="name">POS3</field>
-        </record>
```
yelizariev: Why?

Ramil-Mukhametzyanov: The approach of assigning multi-session to POS after installation is to add each company's POS to the company's single multi-session.
If we can make an exception for POS3, we should give a unique name.

Ramil-Mukhametzyanov (changes requested):
If I activate multi-company, create two POSs for each company, and then try to install pos_multi_session (not from "Your company"), an error will occur.

Ramil-Mukhametzyanov (approved):
If multi company is activated and a POS have been created for each company in a database without demo data, there is no error in the database.

pos_multi_session/demo/demo.xml
Ramil-Mukhametzyanov: @GabbasovDinar you can revert this changes.

Ramil-Mukhametzyanov (changes requested):
Rename the PR, make a note to the changelog, and update version.
The pull request is about proper installation the pos_multi_session* modules in the multi-company mode if the database already has created POS.

Ramil-Mukhametzyanov (approved):

pos_multi_session/models/pos_multi_session_models.py
```diff
@@ -56,18 +59,34 @@ class PosMultiSession(models.Model):
...
+                # Create default multi-session for current company
+                stock_location = self.env['stock.warehouse'].search([('company_id', '=', company.id)], limit=1).lot_stock_id
+                multi_session = self.create({
+                    'name': 'Default Multi Session (%s)' % company.name,
+                    'multi_session_active': False,
+                    'company_id': company.id,
+                    'stock_location_id': stock_location.id
+                })
+                for c in configs:
+                    c.write({
```
yelizariev, line 87: ``configs.write(...`` or  add a comment why it doesn't work this way

yelizariev (changes requested):
Also, check Updating module

GabbasovDinar:
@Ramil-Mukhametzyanov Can you check updating of the module?

PR #995, opened 2019-05-15
10.0 pos wechat rebasing

KolushovAlexandr:


yelizariev:
10.0 branch is low priority. We may continue this on customer request

trojikman:
> 10.0 branch is low priority. We may continue this on customer request

Can I close it now?

PR #1003, opened 2019-05-23
:heart_eyes: show the autostart_longpolling setting after install Multi Session module

GabbasovDinar:


yelizariev (changes requested):
What if we have two modules that use pos_longpolling as dependency? They both add the field? What you can do is keep field in pos_longpolling, but hide it there. Then you disable hiding in pos_multi_session

GabbasovDinar:
@yelizariev I don't understand what do you mean. Only pos_multi_session module add the new field

KolushovAlexandr:
@GabbasovDinar He means to put `invisible="1"` attribute to the `autostart_longpolling` field in pos_longpolling and then make it visible again in pos_multi_session module

KolushovAlexandr:
superseded by https://github.com/it-projects-llc/pos-addons/pull/1044

PR #1005, opened 2019-05-25
:ambulance: error with orders have same numberspaid offline and onlin…

KolushovAlexandr:
…e on different POSes

### DIFF
--- a/pos_multi_session/doc/changelog.rst
+++ b/pos_multi_session/doc/changelog.rst
@@ -1,3 +1,8 @@
+`4.2.8`
+-------
+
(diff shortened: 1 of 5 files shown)


Ramil-Mukhametzyanov (approved):

pos_multi_session/doc/changelog.rst
```diff
@@ -1,3 +1,8 @@
+`4.2.8`
+-------
+
+- **Fix:** Error with orders have the same numbers were paid offline and online on different POSes
```
yelizariev: How it works now? What will happened in the described scenario?

yelizariev (changes requested):

yelizariev (approved):

PR #1018, opened 2019-05-30
:ambulance: cannot set status of active daemon

KolushovAlexandr:


hw_printer_network/controllers/hw_printer_network_controller.py
```diff
@@ -81,6 +82,15 @@ def get_network_printer(self, ip, name=None):
...
         return None
 
+    def push_task(self, task, data=None):
+        if self.on_task_pushing:
+            time.sleep(1)
+            self.push_task(task, data)
+            return
+        self.on_task_pushing = True
+        super(EscposNetworkDriver, self).push_task(task, data)
+        self.on_task_pushing = False
```
yelizariev, line 92: Lock objects are invented for this https://docs.python.org/3/library/threading.html

```
acquire(blocking=True, timeout=-1)¶

When invoked with the blocking argument set to True (the default), block until the lock is unlocked, then set it to locked and return True.
```

yelizariev (changes requested):

Ramil-Mukhametzyanov (approved):
Our customer has passed many POS test and approved in the actual environment.

yelizariev:
@Ramil-Mukhametzyanov thanks for information, but current technical implementation looks naive

em230418 (changes requested):
Agree with Ivan https://github.com/it-projects-llc/pos-addons/pull/1018/files#r291501570

PR #1027, opened 2019-06-17
10.0 ms

GabbasovDinar:


pos_multi_session/models/pos_multi_session_models.py
```diff
@@ -41,6 +43,20 @@ def _search_current_session_state(self, operator, value):
...
+        res = super(PosConfig, self).open_ui()
+        active_sessions = self.env['pos.session'].search(
+            [('state', '!=', 'closed'), ('config_id.multi_session_id', '=', self.multi_session_id.id)])
+        if len(active_sessions) == 1 and active_sessions.id == self.current_session_id.id and self.multi_session_id.load_unpaid_orders:
+            orders = self.multi_session_id.get_unpaid_ms_orders()
+            if orders:
+                orders.write({
+                    'state': 'draft',
+                    'run_ID': self.multi_session_id.run_ID
+                })
```
yelizariev, line 57: it works only with same-server deployment

pos_multi_session/models/pos_multi_session_models.py
```diff
@@ -41,6 +43,20 @@ def _search_current_session_state(self, operator, value):
...
 
+    @api.multi
+    def open_ui(self):
+        res = super(PosConfig, self).open_ui()
+        active_sessions = self.env['pos.session'].search(
+            [('state', '!=', 'closed'), ('config_id.multi_session_id', '=', self.multi_session_id.id)])
+        if len(active_sessions) == 1 and active_sessions.id == self.current_session_id.id and self.multi_session_id.load_unpaid_orders:
+            orders = self.multi_session_id.get_unpaid_ms_orders()
+            if orders:
+                orders.write({
```
yelizariev, line 54: The code is difficult to read. You must add some comments about what is going on here

yelizariev (changes requested):

Ramil-Mukhametzyanov (approved):
:+1: It will be very helpful to recover losses after order conflict

PR #1070, opened 2019-08-02
11 fix duplicated event listener invoice pay

ssaid:
Duplicate event listener when pos_keyboard and pos_invoice_pay are installed.


yelizariev (approved):
Looks ok at first glance

Ramil-Mukhametzyanov (commented):
Merge is here https://github.com/it-projects-llc/pos-addons/pull/1126

PR #1074, opened 2019-08-09
:shield: switching to POS menu could take time.

yelizariev:
See https://travis-ci.com/it-projects-llc/pos-addons/jobs/223473947


yelizariev:
I approve to merge it now

PR #1087, opened 2019-08-21
12.0 pos mobile fix clientlist

em230418:
Reason is that `$('.pos.mobile .window').css({display: 'table'})` from `change_screen_type` was not setting style for all elements in selector due to some bug in jquery 1.11.1. But somehow
`$('.pos.mobile .window').css({display: 'none'})` is working correct, so it works correctly without this css rule.

### DIFF
--- a/pos_mobile/README.rst
+++ b/pos_mobile/README.rst
@@ -10,6 +10,7 @@ Credits
 Contributors
 ------------
 * `Dinar Gabbasov <https://it-projects.info/team/GabbasovDinar>`__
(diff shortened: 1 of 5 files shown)


Ramil-Mukhametzyanov (changes requested):
There is no vertical scrolling in the customer list screen.

Ramil-Mukhametzyanov (approved):

yelizariev (approved):

em230418:
I approve to merge it now

PR #1116, opened 2019-10-10
:arrow_up::one::three: - pos_mobile 

fedoranvar:


trojikman:
Moved to the private repo because dependency `pos_debranding` is ported in private repo now

PR #1129, opened 2019-10-19
:x: Directories with unsupported modules removed

NigmatullinAN:


trojikman:
It's decided to don't merge branches, so next version (odoo 14) will be empty at the beginning at anyway

PR #1150, opened 2019-11-01
:shield: travis: use python3.6

yelizariev:
It's a try to fix encoding problem on printing HTML in random tour errors

https://api.travis-ci.com/v3/job/212575529/log.txt
'ascii' codec can't encode character u'\u20ac' in position 7: ordinal not in range(128)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: python
 
 python:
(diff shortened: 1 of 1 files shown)


yelizariev:
I approve to merge it now

PR #1160, opened 2019-11-18
[10.0] pos_debt_notebook: fix some bugs

alexis-via:
1) create account of debt journal with reconcile=True
2) When we have demo data, don't create 2 products with same default_code
3) Partial fix for button "Pay Full Debt" on customer list in POS frontend. The feature triggered by a click on that button is still buggy: in some scenarios, the payment line is not created.

### DIFF
--- a/pos_debt_notebook/data/demo.xml
+++ b/pos_debt_notebook/data/demo.xml
@@ -10,7 +10,7 @@
 
         <record id="product_credit_product" model="product.product">
             <field name="name">Credits top-up</field>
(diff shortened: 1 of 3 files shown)


alexis-via:
One thing that I noticed: the button "Pay Full Debt" on customers list is sometimes buggy, but the button "Pay Full Debt" on the payment screen works fine. Note that the code is not the same for those 2 buttons.

Ramil-Mukhametzyanov (commented):
@alexis-via 
Thank you for the pull-request. We'll check it.

PR #1184, opened 2019-12-10
porting product_lot 10.0 -> 11.0

Enigma228322:
Signed-off-by: Vildan Safin <vildan.safin.99@gmail.com>

### DIFF
--- a/pos_cashbox/__manifest__.py
+++ b/pos_cashbox/__manifest__.py
@@ -6,7 +6,7 @@
     "category": "Point of Sale",
     # "live_test_url": "http://apps.it-projects.info/shop/product/pos-cashbox?version=10.0",
     "images": ["images/pos_cashbox_main.png"],
-    "version": "10.0.1.0.0",
+    "version": "11.0.1.0.0",
     "application": False,
     "author": "IT-Projects LLC, Dinar Gabbasov",
     "support": "pos@it-projects.info",
--- a/pos_debt_notebook/static/src/js/pos.js
+++ b/pos_debt_notebook/static/src/js/pos.js
@@ -4,7 +4,7 @@
  * Copyright 2017-2018 Gabbasov Dinar <https://it-projects.info/team/GabbasovDinar>
  * Copyright 2018 Kolushov Alexandr <https://it-projects.info/team/KolushovAlexandr>
  * License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html). */
-
+/* eslint complexity: "off"*/
 odoo.define("pos_debt_notebook.pos", function(require) {
     "use strict";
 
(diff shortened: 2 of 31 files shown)


Enigma228322:
@yelizariev @Ramil-Mukhametzyanov pr needs review

product_lot/doc/index.rst
```diff
@@ -8,7 +8,8 @@ Installation
 
 Configuration
 =============
-Activate product variants in Sales app settings.
+* Open ``Inventory >> Settings``
+Activate product ``Attributes and Variants``.
```
KolushovAlexandr: * wrong indent
* activate **product**??

product_lot/views.xml
```diff
@@ -1,5 +1,10 @@
 <?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright 2014-2017 Ivan Yelizariev <https://it-projects.info/team/yelizariev>
+     Copyright 2017-2018 Artyom Losev <https://github.com/ArtyomLosev>
+     License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl.html). -->
+
 <openerp>
```
KolushovAlexandr: `<odoo>`

product_lot/models.py
```diff
@@ -1,3 +1,9 @@
+# Copyright 2014-2017 Ivan Yelizariev <https://it-projects.info/team/yelizariev>
+# Copyright 2019 Vildan Safin <https://it-projects.info/team/yelizariev>
```
KolushovAlexandr: sorting done wrong. From first to last ones

KolushovAlexandr (changes requested):
qty on hand is not updated after `split lot` button clicked

product_lot/views.xml
```diff
@@ -1,19 +1,23 @@
...
-                </xpath>
-            </field>
-        </record>
-</openerp>
+<!-- Copyright 2014-2017 Ivan Yelizariev <https://it-projects.info/team/yelizariev>
+     Copyright 2017-2018 Artyom Losev <https://github.com/ArtyomLosev>
+     License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl.html). -->
+
+<odoo>
+    <record id="product_lot_normal_form_view" model="ir.ui.view">
```
KolushovAlexandr: Keep commit history clean. Do not change what may not be changed. Leave indents for `<record>...` as they were

KolushovAlexandr (changes requested):
* Qty on hand is updated after `splitting lot`.
* Forecasted qty is updated twice
For example, if we have 5 'computer case' in 'ice cream' (on hand and forecasted for both of them are 0) then after splitting the lot we have
*  -1 forecasted for 'ice cream'
* 5 on hand for 'computer case'
* 10 forecasted for 'computer case'

@ilmir-k I'm not sure about the first option Is it how it should be?

@Enigma228322 
forecasted should not be updated twice after `splitting`

PR #1205, opened 2020-02-12
Fixing lints in v10.0

Enigma228322:


yelizariev:
I'll disable that check globally

PR #1256, opened 2020-06-01
:shield: exclude hw_escpos from testing

trojikman:


trojikman:
I approve to merge it now

PR #1257, opened 2020-06-02
:shield: exclude hw_printer_network from testing

trojikman:


trojikman:
I approve to merge it now

PR #1188, opened 2019-12-18
12.0

ejbdev:
Changes:
Module: `pos_orderline_absolute_discount`
changes introduced by commit [here](https://github.com/it-projects-llc/pos-addons/pull/1188/commits/a227a542bbd6c9da246c0b5561ee93900ec6048d) prevent key error when template [report_saledetails](https://github.com/odoo/odoo/blob/dbc664e5d5f28d6fdd8f791ee349a842fd7d22d0/addons/point_of_sale/views/report_saledetails.xml#L69-L72) is render.

Module: `pos_longpolling`
changes introduced by commit [here](https://github.com/it-projects-llc/pos-addons/pull/1188/commits/45edeae5f0b893f9517e9f563d626e16ed715d27) prevent parse error when the channel name is string, e.g. `my_channel`

PR #1260, opened 2020-06-08
12.0 pos longpolling

trojikman:
moved from #1188 and cleaned


trojikman:
I approve to merge it now

PR #1139, opened 2019-10-23
:arrow_up::one::three: pos_debt_notebook

fedoranvar:

PR #1263, opened 2020-06-15
:arrow_up::one::three: pos_debt_notebook

trojikman:
From #1139 by @fedoranvar 
There were difficulties with fixing merge conflicts


trojikman:
Есть недоработки, так что WIP

PR #1283, opened 2020-06-29
:shield: hw_escpos's dependency for #1281

trojikman:


trojikman:
I approve to merge it now

PR #1288, opened 2020-07-10
:book: change module version after #1253

trojikman:


trojikman:
I approve to merge it now

PR #1295, opened 2020-09-07
:book: update usage instruction

trojikman:


trojikman:
I approve to merge it now

PR #1301, opened 2020-10-02
pos_logout

almas50:


trojikman (changes requested):
13.0 version of Odoo has a built-in ability to log out, so in this module you can now leave only the option to set auto log out timer.
I described the problem in more detail in the appropriate telegram group

trojikman (approved):
Thanks for the job, it looks good!

trojikman:
I approve to merge it now

PR #1304, opened 2020-10-08
Add config boolean to force mobile view always

EstebanMonge:


pos_mobile/views/pos_mobile_view.xml
```diff
@@ -17,6 +17,17 @@
...
                     </div>
                 </div>
+                <div class="col-xs-12 col-md-6 o_setting_box">
+                    <div class="o_setting_left_pane">
+                        <field name="force_mobile" />
+                    </div>
+                    <div class="o_setting_right_pane">
+                        <label for="force_mobile" />
+                        <div class="text-muted">
+                                Force switch to Mobile view (depends of Auto Mobile View)
```
yelizariev, line 27: What means *depends of Auto Mobile View*?

yelizariev (commented):
Good in general

pos_mobile/views/pos_mobile_view.xml
EstebanMonge: When is checked Auto Mobile View, this check forces Moble View always, for example if you have a desktop computer but wants to have mobile view.

yelizariev: We could hide this option if auto_mobile is off. This way is more user friendly

PR #1307, opened 2020-10-21
11.0 POS multi session: restore unpaid orders

Enigma228322:


Enigma228322:
I'll fix tour errors later

Issue #1308, opened 2020-11-04
pos_invoice_pay: extra items in the printed receipt


trojikman:
Impacted versions: 10.0 at least
Source: https://github.com/itpp-labs/pos-addons/tree/10.0/pos_invoice_pay

Steps to reproduce:
1) Make steps to pay sale order or Invoice due to the docs
2) When you try to print a check you get something like that:
![invoice_receipt](https://user-images.githubusercontent.com/23641186/98139815-72119b80-1ee6-11eb-8e97-8957523048cf.png)
As you can see there are unnecessary items that I would not like to see in the receipt.

Expected behavior:
get a normal check without unnecessary elements in it

PR #1311, opened 2020-11-11
:ambulance: elements when printing a receipt through a browser

trojikman:
Solves #1308


trojikman:
I approve to merge it now

PR #1336, opened 2020-12-10
:ambulance: make pos_notes field visible in product form

trojikman:


trojikman:
I approve to merge it now

PR #1374, opened 2021-02-18
:euro: make the module free

trojikman:
if you need support or need us to fix a bug take a look at our subscription options https://www.patreon.com/itpp


trojikman:
I approve to merge it now

PR #1398, opened 2021-03-23
:book::one::two: update versions in docs after migration

trojikman:


trojikman:
I approve to merge it now

PR #1422, opened 2021-07-06
:rainbow: replace <openerp> tag with <odoo>

yelizariev:
to avoid problems in migration scripts


yelizariev:
I approve to merge it now

PR #1477, opened 2022-09-14
[FIX] POS order return impossible (pos_orderline_absolute_discount)

vincent-mundo:
Hello, here is a pull request to validate an article return with or without "absolute_discount".

At the beginning the return of article was possible only when there was an absolute_discount (discount by number, not percentage).

Now there is the return no matter what.
I tested it in all possibilities, from ticket to invoice.

Thanks

### DIFF
--- a/pos_orderline_absolute_discount/models/pos_order_model.py
+++ b/pos_orderline_absolute_discount/models/pos_order_model.py
@@ -53,6 +53,8 @@ class PosOrderLine(models.Model):
                 if fpos
                 else line.tax_ids
             )
+            price = line.price_unit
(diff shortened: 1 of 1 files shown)


pos_orderline_absolute_discount/models/pos_order_model.py
```diff
@@ -62,12 +62,13 @@ def _compute_amount_line_all(self):
                     product=line.product_id,
                     partner=line.order_id.partner_id,
                 )
-                line.update(
+                res.update(
```
yelizariev, line 65: it should be `line` I think

flotho:
thanks @yelizariev , we're going to have a look to this with @gaelTorrecillas

flotho:
I think we can close this one because it was not totally correct.
The #1485 is really better @yelizariev

PR #1485, opened 2023-02-15
fix method compute amound line all

JulienMartinez:
Problem with returning an item from the pos which was impossible causing an error message .

Fix that replaces Pull request #1477 :
- https://github.com/itpp-labs/pos-addons/pull/1477

### DIFF
--- a/pos_orderline_absolute_discount/models/pos_order_model.py
+++ b/pos_orderline_absolute_discount/models/pos_order_model.py
@@ -45,29 +45,30 @@ class PosOrderLine(models.Model):
         "price_unit", "tax_ids", "qty", "discount", "product_id", "absolute_discount"
     )
     def _compute_amount_line_all(self):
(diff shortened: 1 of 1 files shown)


flotho (approved):
code review + tests

flotho:
ping @gaelTorrecillas @yelizariev

yelizariev (commented):
I guess we should delete `api.depends`

api.depends was used in v11:

https://github.com/odoo/odoo/blob/43200e9dcc1d91cfd5c1e5e939b27e713d20cc7c/addons/point_of_sale/models/pos_order.py#L1075-L1085


But then it was refactored. Code for v14:
https://github.com/odoo/odoo/blob/9184dba02e7177e7139649b56e73c3857f4d008b/addons/point_of_sale/models/pos_order.py#L826-L832

